PR #13, opened 2024-10-25
feat: add SBOM CRD database storage

fabriziosestito:
This PR adds the SBOM CRD to the database storage.
It implements the [apiserver storage interface](https://danielmangum.com/posts/k8s-asa-the-storage-interface/) so that resources can be stored in an external database.
As for now, only sqlite is supported, but it could be extended to other databases by changing the driver.

We are using [modernc.org/sqlite](https://pkg.go.dev/modernc.org/sqlite) which does not require CGO enabled.
The watch is implemented by using [watch.Broadcaster](https://pkg.go.dev/k8s.io/apimachinery/pkg/watch#Broadcaster).

The store can be generalized by using generics to support the other CRD provided by the storage component (Image, VulnerabilityReport).


helm/templates/controller/role.yaml
```diff
@@ -19,45 +20,21 @@ rules:
...
-  - watch
-- apiGroups:
-  - sbombastic.sbombastic.rancher.io
-  resources:
-  - images/finalizers
-  verbs:
-  - update
-- apiGroups:
-  - sbombastic.sbombastic.rancher.io
-  resources:
```
flavio, line 53: I'm just confused, why these rules have been dropped?

internal/controller/image_controller.go
```diff
@@ -38,9 +38,9 @@ type ImageReconciler struct {
 	Publisher messaging.Publisher
 }
 
-// +kubebuilder:rbac:groups=sbombastic.sbombastic.rancher.io,resources=images,verbs=get;list;watch;create;update;patch;delete
-// +kubebuilder:rbac:groups=sbombastic.sbombastic.rancher.io,resources=images/status,verbs=get;update;patch
-// +kubebuilder:rbac:groups=sbombastic.sbombastic.rancher.io,resources=images/finalizers,verbs=update
+// +kubebuilder:rbac:groups=sbombastic.rancher.io,resources=images,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=sbombastic.rancher.io,resources=images/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=sbombastic.rancher.io,resources=images/finalizers,verbs=update
```
flavio, line 43: this is what led to the rbac rules to be dropped

internal/storage/sbom_store.go
```diff
@@ -0,0 +1,525 @@
...
+
+	options := &generic.StoreOptions{RESTOptions: optsGetter, AttrFunc: GetAttrs}
+	if err := store.CompleteWithOptions(options); err != nil {
+		return nil, fmt.Errorf("unable to complete store with options: %w", err)
+	}
+
+	return store, nil
+}
+
+var _ storage.Interface = &sbomStore{}
```
flavio, line 59: What is the purpose of that?

internal/storage/sbom_store.go
```diff
@@ -0,0 +1,525 @@
...
+	if name == "" || namespace == "" {
+		return storage.NewInternalErrorf("invalid key: %s", key)
+	}
+
+	tx, err := s.db.BeginTxx(ctx, nil)
+	if err != nil {
+		return storage.NewInternalError(err.Error())
+	}
+	defer func() {
+		if err := tx.Rollback(); !errors.Is(err, sql.ErrTxDone) {
```
flavio, line 140: So we always try to roolback, but if the transaction was successfully committed we get `sql.ErrTxDone` and we just ignore that, because it's all good. Do I get it right?

internal/storage/sbom_store.go
```diff
@@ -0,0 +1,525 @@
...
+
+	if err := s.broadcaster.Action(watch.Deleted, out); err != nil {
+		return storage.NewInternalError(err.Error())
+	}
+
+	return nil
+}
+
+// Watch begins watching the specified key. Events are decoded into API objects,
+// and any items selected by 'p' are sent down to returned watch.Interface.
```
flavio, line 185: I think `p` is now gone from the implementation of this function. Is it `opts` maybe?

internal/storage/sbom_store.go
```diff
@@ -0,0 +1,525 @@
...
+// If resource version is "0", this interface will get current object at given key
+// and send it in an "ADDED" event, before watch starts.
+func (s *sbomStore) Watch(ctx context.Context, key string, opts storage.ListOptions) (watch.Interface, error) {
+	if opts.ResourceVersion == "" {
+		return s.broadcaster.Watch()
+	}
+
+	if opts.ResourceVersion == "0" {
+		sbom := &v1alpha1.SBOM{}
+		if err := s.Get(ctx, key, storage.GetOptions{}, sbom); err != nil {
```
flavio, line 198: We have to take into account the `opts`, maybe the user is doing a watch using a label selector or similar

internal/storage/sbom_store.go
```diff
@@ -0,0 +1,525 @@
...
+			return nil, err
+		}
+
+		return s.broadcaster.WatchWithPrefix([]watch.Event{{Type: watch.Added, Object: sbom}})
+	}
+
+	sbomList := &v1alpha1.SBOMList{}
+	if err := s.GetList(ctx, key, storage.ListOptions{}, sbomList); err != nil {
```
flavio, line 206: We have to take into account the `opts`, maybe the user is doing a watch using a label selector or similar

internal/storage/sbom_schema.go
```diff
@@ -0,0 +1,20 @@
+package storage
+
+const CreateSBOMTableSQL = `
+CREATE TABLE IF NOT EXISTS sboms (
+    id INTEGER PRIMARY KEY,
+    name VARCHAR(253) NOT NULL,
+    namespace VARCHAR(253) NOT NULL,
+    object TEXT NOT NULL,
+    UNIQUE(name, namespace)
+);
```
flavio, line 10: Based on what I've seen below, we put the json encoded object (the whole CRD) into the `object` column.

I think we have to extract the labels from the object and store them into dedicated columns. Users can do `GET`, `DELETE`, `WATCH` operations using label selectors

internal/storage/sbom_store.go
```diff
@@ -0,0 +1,525 @@
...
+
+	return s.broadcaster.WatchWithPrefix(events)
+}
+
+// Get unmarshals object found at key into objPtr. On a not found error, will either
+// return a zero object of the requested type, or an error, depending on 'opts.ignoreNotFound'.
+// Treats empty responses and nil response nodes exactly like a not found error.
+// The returned contents may be delayed, but it is guaranteed that they will
+// match 'opts.ResourceVersion' according 'opts.ResourceVersionMatch'.
+func (s *sbomStore) Get(ctx context.Context, key string, opts storage.GetOptions, objPtr runtime.Object) error {
```
flavio, line 222: It looks like we are not taking into account the selectors that might be provided by the user

internal/storage/sbom_store.go
```diff
@@ -0,0 +1,525 @@
...
+		}
+		*out = *sbom
+
+		if err := s.broadcaster.Action(watch.Modified, updatedSBOM); err != nil {
+			return storage.NewInternalError(err.Error())
+		}
+
+		if err := tx.Commit(); err != nil {
+			return storage.NewInternalError(err.Error())
+		}
```
flavio, line 449: Shouldn't we invert the order:

1. commit to db
2. if everything went fine (the record is into the db) -> notify the watchers

internal/storage/sbom_store.go
```diff
@@ -0,0 +1,525 @@
...
+// TODO: Remove when storage.Interface will be separate from etc3.store.
+// Deprecated: Added temporarily to simplify exposing RequestProgress for watch cache.
+func (s *sbomStore) RequestWatchProgress(_ context.Context) error {
+	// As this is a deprecated method, we are not implementing it.
+	return nil
+}
+
+// extractNameAndNamespace extracts the name and namespace from the key.
+// Used for single object operations.
+func extractNameAndNamespace(key string) (string, string) {
```
flavio, line 504: Can you put a comment about how the typical key looks like?

internal/storage/sbom_store_test.go
```diff
@@ -0,0 +1,348 @@
...
+	suite.Run(t, &sbomStoreTestSuite{})
+}
+
+func (suite *sbomStoreTestSuite) TestCreate() {
+	sbom := &v1alpha1.SBOM{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "test",
+			Namespace: "default",
+		},
+	}
```
flavio, line 58: Silly requirement, can you add also some labels, annotations and the `spec`? I know the `Spec` is going to change soon, but at least we can test the whole object

internal/storage/sbom_store_test.go
```diff
@@ -0,0 +1,348 @@
...
+	out := &v1alpha1.SBOM{}
+	err := suite.store.Create(context.Background(), key, sbom, out, 0)
+	suite.Require().NoError(err)
+
+	suite.EqualValues(sbom, out)
+	suite.Equal("1", out.ResourceVersion)
+}
+
+func (suite *sbomStoreTestSuite) TestDelete() {
+	sbom := &v1alpha1.SBOM{
```
flavio, line 70: It would be great to test also the delete using label selectors, like delete all the SBOMs that have label `foo` equal `bar` and are located into the namespace `default`.

Also, delete all the SBOMs with label `foo=bar`, regardless of their namespace

internal/storage/sbom_store_test.go
```diff
@@ -0,0 +1,348 @@
...
+
+	events := collectEvents(watcher)
+	suite.Require().Len(events, 2)
+	suite.Equal(watch.Added, events[0].Type)
+	suite.Equal(sbom, events[0].Object)
+	suite.Equal(watch.Deleted, events[1].Type)
+	suite.Equal(sbom, events[1].Object)
+}
+
+func (suite *sbomStoreTestSuite) TestWatchSpecificResourceVersion() {
```
flavio, line 170: Not related with this test, just picking a "random" line. I think we should also test watches that are using selectors

internal/storage/sbom_store_test.go
```diff
@@ -0,0 +1,348 @@
...
+				input.(*v1alpha1.SBOM).Spec.Data.Raw = []byte(`{"foo": "bar"}`)
+
+				return input, ptr.To(uint64(0)), nil
+			}
+			updatedSBOM := &v1alpha1.SBOM{}
+			err := suite.store.GuaranteedUpdate(context.Background(), test.key, updatedSBOM, test.ignoreNotFound, test.preconditions, tryUpdate, nil)
+
+			if test.expectedError != nil {
+				suite.Require().Error(err)
+				suite.Require().Equal(test.expectedError.Error(), err.Error())
```
flavio, line 290: should we also check that the db contents didn't change?

flavio (commented):
Outstanding work! I left some comments

helm/templates/controller/role.yaml
fabriziosestito: because they were wrong: 
`sbombastic.sbombastic.rancher.io` instead of `sbombastic.rancher.io`

internal/storage/sbom_store.go
fabriziosestito: yes this is a common pattern, see: https://go.dev/doc/database/execute-transactions

The difference is that we check the error returned

internal/storage/sbom_store.go
fabriziosestito: This is the original document from the Kubernetes authors. I will fix this in our codebase and try to fix upstream too. (p was referring to SelectionPredicate)

internal/storage/sbom_store_test.go
fabriziosestito: This feature does not exist. You are describing a combination of a filtered `GetList` and a `Delete`.

internal/storage/sbom_store.go
fabriziosestito: This happens after this block where we do a `GetList` passing the options.|
The interface documentation states:

`// If resource version is "0", this interface will get current object at given key
// and send it in an "ADDED" event, before watch starts.`

Which means that we are using a `Get` call which doesn't have label predicates.

internal/storage/sbom_store.go
fabriziosestito: The `Get` call doesn't use selectors.

internal/storage/sbom_store.go
fabriziosestito: You are right

internal/storage/sbom_store.go
fabriziosestito: This convention ensures that the code will not compile if we do not implement the `storage.Interface` correctly.

internal/storage/sbom_schema.go
fabriziosestito: `GET` and `DELETE` operations do not support label selectors directly.

As per the `GetList`, which is also used by the `WATCH`, we are getting all the objects under a certain key, and then we are applying the label predicate programmatically. See https://github.com/fabriziosestito/sbombastic/blob/e0fd7b22decd0fe97054d45f6d958b31ffb14570/internal/storage/sbom_store.go#L294

I've tracked this as debt here  #16 and as a feature here: #15

flavio (approved):
LGTM, ship it :)

Issue #14, opened 2024-10-30
Generic store implementation


fabriziosestito:
The storage component will manage `Image`, `SBOM`, and `VulnerabilityReport` records within the database. Currently, only the SBOM store is implemented. 
Since each CRD follows the same behavior, the store implementation can be generalized. To support multiple types, we should refactor the SBOM store as a generic implementation:

```go
type genericStore[T runtime.Object] {
  db *sqlx.DB
  tableName string
}
```

PR #18, opened 2024-10-31
refactor(storage): implement generic store

fabriziosestito:
This PR refactors the existing `sbomStorage` to a generic store that can be used with any `runtime.Object`.
Some reflection was needed to make the code generic, similar to what was done in the [etcd3 store of Kubernetes](https://github.com/kubernetes/apiserver/blob/master/pkg/storage/etcd3/store.go).

Fixes: #14


internal/storage/store.go
```diff
@@ -0,0 +1,540 @@
...
+		return err
+	}
+
+	for _, objectRecord := range objectRecords {
+		obj := s.newFunc()
+		if err := json.Unmarshal(objectRecord.Object, obj); err != nil {
+			return storage.NewInternalError(err.Error())
+		}
+
+		ok, err := opts.Predicate.Matches(obj)
```
flavio, line 276: Should we plan to use SQL selectors in the future to filter the results at the db level instead of using this approach? We don't have to fix it right now, just to track this "technical debt"

flavio (approved):
LGTM. You can merge it

internal/storage/store.go
flavio: I take it back, it's already tracked

PR #25, opened 2024-11-07
feat: move Image to storage and implement field selectors

fabriziosestito:
This PR moves the `Image` type from the controller to storage and implements image metadata field selectors, removing the labels.
It also implements an `Image` store.


helm/templates/worker/deployment.yaml
```diff
@@ -15,7 +15,7 @@ spec:
       labels:
         worker: "true"
     spec:
-      serviceAccountName: apiserver
+      serviceAccountName: controller-manager
```
fabriziosestito, line 18: @flavio we need to change this when we work on the helm chart

flavio (approved):
What a PR, that was huge :sweat_smile: 

Everything LGTM, I left some minor comments

PR #28, opened 2024-11-13
feat: SBOM generation

fabriziosestito:
This PR implements the SBOM generation handler in the worker component.

When receiving a `GenerateSBOM` message, the worker generates a SBOM in SPDX-JSON format and stores it in the SBOM resource.
For now, we are using trivy to create the SBOMs, but we plan to use multiple adapters in the future.

A few tweaks to the storage types were needed.


internal/handlers/generate_sbom.go
```diff
@@ -0,0 +1,109 @@
...
+	)
+
+	sbomFile, err := os.CreateTemp("/tmp", "trivy.sbom.*.json")
+	if err != nil {
+		return fmt.Errorf("failed to create temporary file: %w", err)
+	}
+	defer func() {
+		if err := sbomFile.Close(); err != nil {
+			h.logger.Error("failed to close temporary file", zap.Error(err))
+		}
```
flavio, line 63: nit: I don't think we need to care about closing the file. Do we?

internal/handlers/generate_sbom_test.go
```diff
@@ -0,0 +1,88 @@
...
+			Namespace: "default",
+		},
+		Spec: storagev1alpha1.ImageSpec{
+			ImageMetadata: storagev1alpha1.ImageMetadata{
+				Registry:    "docker",
+				RegistryURI: "docker.io",
+				Repository:  "busybox",
+				Tag:         "1.37.0",
+				Platform:    "linux/amd64",
+				Digest:      "sha256:123",
```
flavio, line 37: The test is actually running trivy and building a real sbom of this image. Do I get it right?

Some questions:
- Should we point to an image we have control (this version of busybox is going to disappear from the docker registry at some point in time)
- Should we create right now an interface for the sbom generator? This would allow us to use a mock for trivy right now and later to have a grype based sbom generator

flavio (commented):

internal/handlers/generate_sbom.go
fabriziosestito: this is needed to prevent resource leaks

helm/templates/worker/deployment.yaml
```diff
@@ -17,6 +17,12 @@ spec:
...
-        - name: wardle-server
+        - name: worker
           image: {{ .Values.worker.image.repository }}:{{ .Values.worker.image.tag }}
           imagePullPolicy: {{ .Values.worker.image.pullPolicy }}
+          volumeMounts:
+            - mountPath: /tmp
+              name: tmp-volume
+      volumes:
+        - name: tmp-volume
+          emptyDir: {}
```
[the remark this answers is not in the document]
fabriziosestito: I didn't want to put this in memory as it also holds the trivy cache.
As a future improvement, we could expose the cache as a persistent volume so that the cache is persisted.

internal/handlers/generate_sbom_test.go
fabriziosestito: I've tracked both as an issue.

The adapter is out of scope for the MVP, keep in mind that this will only move the integration test on the adapter side.

flavio: I'm fine with that

helm/templates/worker/deployment.yaml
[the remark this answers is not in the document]
flavio: Yes, that's totally fine with me. I think we don't need to have `/tmp` to be an emptyDir, this is already read-write inside of a Pod. Should we instead create a dedicated path for these operations? Like `/var/run/worker/`?

internal/handlers/generate_sbom.go
flavio: Go has a finalizer, when the GC cleans the sbomFile it will also close the file descriptor. But I think it's nice to be explicit about that, it's more "clean"

PR #37, opened 2024-11-18
fix(storage): tryUpdate should only retry index conflict errors

fabriziosestito:
The GuaranteedUpdate function receives a tryUpdate function as a parameter. According to the documentation, the update is retried indefinitely only when an index conflict error occurs. However, we were also retrying indefinitely for other errors, leading to an infinite loop.

This PR fixes that issue.


flavio (approved):
thanks for the fix!

PR #38, opened 2024-11-18
fix(storage): return key exists error instead of bubbling up a SQL error

fabriziosestito:
Return key exists error instead of bubbling up a SQL error.


flavio (approved):
LGTM

PR #41, opened 2024-11-19
fix: improve the logic of handlers to cover scenarios where resources already exist

fabriziosestito:
improve the logic of handlers to cover scenarios where resources already exist

- `CreateCatalogHandler`: skip creation if an `Image` already exists, delete obsolete images.
- `GenerateSBOMHandler`: skip creation if a `SBOM` already exists
- `ScanSBOMHandler`: create or update `VulnerabilityReport`

### DIFF
--- a/internal/handlers/create_catalog.go
+++ b/internal/handlers/create_catalog.go
@@ -18,6 +18,7 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	storagev1alpha1 "github.com/rancher/sbombastic/api/storage/v1alpha1"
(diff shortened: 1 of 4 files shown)


internal/handlers/create_catalog.go
```diff
@@ -75,7 +76,7 @@ func (h *CreateCatalogHandler) Handle(message messaging.Message) error {
 		return fmt.Errorf("cannot discover repositories: %w", err)
 	}
 
-	var imageNames []string
+	newImageNames := sets.Set[string]{}
```
flavio, line 79: nit: `newImageNames` makes me think about the "net new image", the one that have been discovered and are not yet known (as in, do not have an Image CR associated).

Can we pick a different name?

internal/handlers/create_catalog.go
```diff
@@ -259,6 +264,28 @@ func (h *CreateCatalogHandler) transportFromRegistry(registry *v1alpha1.Registry
 	return transport
 }
 
+// deleteObsoleteImages deletes images that are not present in the discovered registry anymore.
+func (h *CreateCatalogHandler) deleteObsoleteImages(ctx context.Context, existingImageNames sets.Set[string], newImagesNames sets.Set[string], namespace string) error {
+	for existingImageName := range existingImageNames {
+		if newImagesNames.Has(existingImageName) {
+			continue
+		}
```
flavio, line 272: nit: isn't there an "intersection"-like method provided by this `Set` struct?

flavio (commented):
LGTM, I left two minor comments

internal/handlers/create_catalog.go
fabriziosestito: We could use `sets.Difference` but it doesn't change much as we need to cycle to delete the images anyway. Also, this way we will cycle twice.

Issue #33, opened 2024-11-14
Use slog for logging


fabriziosestito:
Replace `zap` and `logr.Logger` with `slog`.

`logr` will still remain as an indirect dependency of `controller-runtime` since it is used internally.

PR #44, opened 2024-11-22
chore: switch to slog

fabriziosestito:
Change logger from `zap` to `slog`.
Note that we are still using `slog` through `logr` in the controller since this is needed by `controller-runtime`.

Fixes: #33

### DIFF
--- a/cmd/worker/main.go
+++ b/cmd/worker/main.go
@@ -2,13 +2,11 @@ package main
 
 import (
 	"context"
-	"fmt"
+	"log/slog"
 	"net/http"
 	"os"
 	"os/signal"
 	"syscall"
-
-	"go.uber.org/zap"
 
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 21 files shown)


cmd/worker/main.go
```diff
@@ -72,6 +71,6 @@ func main() {
 
 	err = subscriber.Run(ctx)
 	if err != nil {
-		logger.Fatal("Error running worker subscriber", zap.Error(err))
+		logger.Error("Error running worker subscriber", "error", err)
```
flavio, line 74: You forgot to do `os.Exit(1)`

cmd/worker/main.go
```diff
@@ -23,32 +21,33 @@ import (
...
 	config := ctrl.GetConfigOrDie()
 	scheme := scheme.Scheme
 	if err := v1alpha1.AddToScheme(scheme); err != nil {
-		logger.Fatal("Error adding v1alpha1 to scheme", zap.Error(err))
+		logger.Error("Error adding v1alpha1 to scheme", "error", err)
+		os.Exit(1)
 	}
 	if err := storagev1alpha1.AddToScheme(scheme); err != nil {
-		logger.Fatal("Error adding storagev1alpha1 to scheme", zap.Error(err))
+		logger.Error("Error adding storagev1alpha1 to scheme", "error", err)
```
flavio, line 45: Shouldn't we do `os.Exit(1)` too?

flavio (commented):
`slog` looks really neat! I welcome this change.

PR #49, opened 2024-11-28
fix(storage): field selectors

fabriziosestito:
- selectable fields were wrong in the `getAttrs` func
- vulnerability report didn't have field label conversion
- vulnerability report test CRD was missing selectable fields


flavio (approved):
thanks for the fixes!

PR #50, opened 2024-11-28
fix(storage): resolve GuaranteedUpdate saving an empty object instead of the updated one

fabriziosestito:
Resolve the issue with GuaranteedUpdate saving an empty object instead of the updated data.
Update test to be more strict.


flavio (approved):
ouch, nice fix

PR #72, opened 2024-12-16
chore: fix kubernetes dependabot group

fabriziosestito:
Add the missing `k8s.io/apiserver` and reorder it like go.mod.


flavio (approved):
thanks for the fix

PR #120, opened 2025-03-12
fix: do not panic when image platform is not set

flavio:
When the scanner finds an OCI image, it should not panic because the platform is not set.

I think the `nil` handling is required, but we should also avoid to look into the details of something like an OCI artifact, given they are out of scope for SBOMbastic.

I'll create a dedicated issue to keep track of that.

### DIFF
--- a/internal/handlers/registry/client.go
+++ b/internal/handlers/registry/client.go
@@ -174,6 +174,9 @@ func (c *client) GetImageDetails(ref name.Reference, platform *cranev1.Platform)
 
 	// ensure platform is always set
 	platform = cfgFile.Platform()
+	if platform == nil {
+		return ImageDetails{}, fmt.Errorf("cannot get platform for %s", ref)
+	}
 
 	layers, err := img.Layers()
 	if err != nil {


pohanhuangtw:
Hi @flavio ,
I agree that nil handling is necessary to avoid panicking when the platform is not set.

However, Platform doesn’t affect SBOM generation (focused on dependencies, not platform-specific details).

So, I suggest to handle nil by either providing an empty platform or safely managing it.

What do you think ?

pohanhuangtw:
Hi @flavio ,
Does it mean registry may contains artifact? and we need to find a way to escape it?
Currently we use platform existence as the condition?
Thanks

pohanhuangtw (approved):
Approve the quick workaround, but we need to find better way to fix this.

pohanhuangtw:
Hi @flavio ,
I can reproduce issue by simply push the txt into the registry.

Which point out to the core issue, sbombastic **_assume the ref always represent the image_** In [ref, err := name.ParseReference(newImageName)](https://github.com/rancher-sandbox/sbombastic/blob/main/internal/handlers/create_catalog.go#L103C3-L103C48)

To resolve this completely, we need to filter out the non-image reference [here](https://github.com/rancher-sandbox/sbombastic/blob/main/internal/handlers/create_catalog.go#L103C3-L103C48)

In that way, we must read the correct image information out of the reference.

what do you think?

cc: @fabriziosestito @williamlin-suse

flavio:
> Which point out to the core issue, sbombastic assume the ref always represent the image In [ref, err := name.ParseReference(newImageName)](https://github.com/rancher-sandbox/sbombastic/blob/main/internal/handlers/create_catalog.go#L103C3-L103C48)

The [Reference](https://pkg.go.dev/github.com/google/go-containerregistry/pkg/name#Reference) object is created after parsing an OCI reference, which is a string that references something inside of a container registry.


> To resolve this completely, we need to filter out the non-image reference [here](https://github.com/rancher-sandbox/sbombastic/blob/main/internal/handlers/create_catalog.go#L103C3-L103C48)
>
> In that way, we must read the correct image information out of the reference.

A failure at the line you pointed implied the given image name is not valid. This is something that is never going to happen in our case, since the image was discovered inside of a registry, hence it's name was already validated both by the client that pushed it and by the registry itself when the payload was received.

To make it more explicit, the code would return an error when trying to parse something like `www.example.com` or `registry.local.lan/hello+world/busybox` or `registry.local.lan/hello:you-might-be-suprised-but-this-is-not-a-valid-tag+because-of-the-plus-symbol`

pohanhuangtw:
I think I understand the issue, will keep working on [issue](https://github.com/orgs/rancher-sandbox/projects/6/views/1?pane=issue&itemId=102821638&issue=rancher-sandbox%7Csbombastic%7C121)

PR #140, opened 2025-03-24
rfc: DiscoveryJob

williamlin-suse:


docs/rfc/0002_discoveryjob_design.md
```diff
@@ -0,0 +1,194 @@
...
+10. May need to add more fields(like ObservedGeneration) to DiscoveryJob.Status for internal handling
+
+------
+
+An RFC is created with `ACCEPTED` as state. After discussion and adapting, if it
+merges, it is merged as `ACCEPTED`. If it is finally rejected or consensus is
+not reached, it is changed to `REJECTED` and merged.
+
+States:
+
```
flavio, line 171: Leftover

```suggestion
```

flavio (changes requested):
First quick pass. I'll do another one soon, I need some time to ponder the solution

flavio:
I've thought about how to tackle this problem. I've some generic observations about the RFC, then I'll make a proposal.

# Feedback about RFC

- I don't think we need to have all these annotations
- I don't see the value of distinguishing between a discovery job that was triggered manually vs one that was triggered by a scheduler. Do you have some user story in mind?
- I like the idea to keep track of the registry specification that was used to do the discovery. This is can help debug issues.
- We cannot use a simple ticker inside of a go-routine to trigger the scheduled discoveries. The controller can run in HA mode, leading to potentially multiple tickers running at the same time. There's a pattern that can be used to address this problem, we have adopted that inside of Kubewarden to handle the TLS certificates rotation. See the [CertReconciler](https://github.com/kubewarden/kubewarden-controller/blob/596315257f952a9572d1384c1fa2cffa9212dcc9/internal/controller/cert_controller.go#L23).

# Proposal

I propose to use an approach similar to the one taken by Kubernetes with the `CronJob` and its child `Job`, which is also what [Argo Workflows](https://github.com/argoproj/argo-workflows) uses with the `CronWorkflow` and the `Workflow` resources.

I don't think we should go as far as to introduce another higher level CRD like `CronJob`/`CronWorkflow`, I would instead start by expanding the `Registry` CRD to be something like

```yaml
apiVersion: scanner.rancher.io/v1alpha1
kind: Registry
metadata:
  name: registry-example
  namespace: default
spec:
  uri: "https://registry-1.docker.io"
  discoveryJob:
    schedule: "0 1 * * *" # I wonder if using the cron syntax isn't too complicated
    suspend: false # bool value, toggle the scheduled discovery
    failedJobsHistoryLimit: 2 # number of RegistryDiscovery objects with failed state to keep
    successfulJobsHistoryLimit: 1 # number of RegistryDiscovery objects with successful state to keep
  status:
  - lastScheduledTime: "2025-03-24T01:30:00Z"
``` 

A controller would then trigger the creation of `RegistryDiscovery` objects whenever needed by the schedule provided by the user.

The `RegistryDiscovery` objects would look like that:

```yaml
apiVersion: scanner.rancher.io/v1alpha1
kind: RegistryDiscovery
metadata:
  name: registry-example-1742868000 # <Registry name>-< unix timestamp>
  namespace: default
  labels:
    sbombastic.rancher.io/completed: "true"
    sbombastic.rancher.io/cron-registry-discovery: registry.local.lan
    sbombastic.rancher.io/phase: Succeeded # Could be "Pending", "Running", "Succeeded", "Failed", "Error" # similar to https://github.com/argoproj/argo-workflows/blob/8098a14a2f5dc134a2e21e5a6e6b23b4b76e0e21/pkg/apis/workflow/v1alpha1/workflow_phase.go#L6-L13
spec:
  registrySpec: <copy of registry.Spec> # this embedding pattern is common within k8s, see for example Deployment and PodSpec, CronJob and JobSpec
  status:
  - startedAt: "2025-03-24T01:30:00Z"
  - finishedAt: "2025-03-24T01:32:00Z"
  - conditions: # we can use conditions to provide feedback about the progress
    - status: "False"
      type: Running
    - status: "True"
      type: Completed     
```

Users can trigger a discovery job by creating a `DiscoverJob` object manually. The name to be used would be completely arbitrary. This is similar to how a user can create a Job instance starting from a CronJob template (instead of waiting for the schedule to kick in).
Later, we could also develop a kubectl plugin that would simplify the creation of that.

### Recap

The user creates a `Repository` object. If the `Repository` doesn't have a discovery schedule specified, our mutating admission controller will put a default one.

A new reconciliation loop, running only on the leader of the controllers, will handle an internal clock and, whenever needed it will create instances of `RegistryDiscovery`.

Another reconciliation loop will monitor the `RegistryDiscover` objects. Whenever one is created it will then add a new job inside of the NATs queue. The job object can simply reference the `RegistryDiscovery` object instance, the worker will then get that from the API server and start the create catalog job.

The worker will keep the `RegistryDiscovery` object up-to-date to keep track of the progress of the catalog and its final outcome.

The reconciliation loop that handles the schedule of `Registry` objects will regularly purge the old `RegistryDiscovery` objects associated with each given Registry to ensure the `failedJobsHistoryLimit` and `successfulJobsHistoryLimit` limits are respected.

### Open questions

How do we ensure the same Registry is not scanned concurrently?

I think we can take advantage of NATs' deduplication capabilities to remove duplicated entries from the queue. This can be useful to avoid a situation like: user triggers a manual scan of registry `foo`, the event is queued because the workers are busy. The cron ticker creates another discovery job for registry `foo`. It's useless to have to jobs queued, we should keep only one of them.
However, what should be done with the 2nd `RegistryDiscovery` object? This would never be updated...


Moreover, this approach wouldn't solve the problem of having one worker processing the `foo` registry, and have another worker pick a discovery job for the very same registry from the queue.

Maybe one possible approach would be to have a Validating Admission Controller that rejects the creation of `RegistryDiscovery` instances if there's already one resource for the same registry that is being processed/queued. However, I'm not super convinced about this approach.

Finally, we could maybe take all the `RegistryDiscovery` jobs the user/system creates. Then use a Kubernetes Lease (one per registry ofc), to ensure only one job is processed at the same time.

docs/rfc/0002_discoveryjob_design.md
```diff
@@ -0,0 +1,196 @@
...
+
+1. On Rancher UI the schedule setting should be configured on the Registry page.
+
+2. When a Registry CR is reconciled, a DiscoveryJob CR named(by default) discovery-job-{registry.name} is created for storing the scheduled discovery status.
+   Before a new scheduled discovery operation is triggered, status in this DiscoveryJob CR is purged first
+
+3. Controller caches the scheduled discovery settings of all registries.
+   When a Registry CR is reconciled, controller updates this cache accordingly.
+   There is a scheduler go-routine wakes up per-minute and checks the scheduled discovery settings of all registries.
```
fabriziosestito, line 115: This could be a problem if multiple replicas are started. 
We don't want the non-leader replicas to run the scheduler.
Fortunately, this is possible by using the low-level [RunnableFunc](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/manager#RunnableFunc) from controller-runtime.

williamlin-suse:
> The user creates a `Repository` object. If the `Repository` doesn't have a discovery schedule specified, our mutating admission controller will put a default one.
1. Does `Repository` object mean `Registry` object here?
2. So we will need a mutating admission controller in sbombastic, correct? (when no discovery schedule specified, put a default discovery schedule and set suspend=true). Do we need to mutate 

> 
> A new reconciliation loop, running only on the leader of the controllers, will handle an internal clock and, whenever needed it will create instances of `RegistryDiscovery`.
Is my understanding correct :
In k8s only the leader controller pod will be reconciling events while the other two will be on standby

I thought the loop runs in every controller but a controller creates instances of `RegistryDiscovery` only when it's the leader?
(with the help of leaderelection.RunOrDie())

> Moreover, this approach wouldn't solve the problem of having one worker processing the `foo` registry, and have another worker pick a discovery job for the very same registry from the queue.
Yes, I think so.

> Maybe one possible approach would be to have a Validating Admission Controller that rejects the creation of `RegistryDiscovery` instances if there's already one resource for the same registry that is being processed/queued. However, I'm not super convinced about this approach.
So annotations on the `Registry` object about the running `RegistryDiscovery` job is not considered?
I think having a Validating Admission Controller makes it too complicate.

> Finally, we could maybe take all the `RegistryDiscovery` jobs the user/system creates. Then use a Kubernetes Lease (one per registry ofc), to ensure only one job is processed at the same time.
Do you mean when a `RegistryDiscovery` job is running for registry-A, no `RegistryDiscovery` job for all registries can run at the same time?
The "two `RegistryDiscovery` jobs cannot run at the same time" limitation is per-registry or for-all-registries ?

docs/rfc/0002_discoveryjob_design.md
williamlin-suse: Could we leverage leaderelection.RunOrDie() on a resourcelock.LeaseLock object considering controller has the k8s RBAC for leases?

fabriziosestito: `RunnableFunc` is the proper way to do so, using leader election would not start the func when the manager has started and we don't want to deal with that.
Please take a look to: https://github.com/kubewarden/kubewarden-controller/blob/596315257f952a9572d1384c1fa2cffa9212dcc9/internal/controller/cert_controller.go#L33

flavio:
> > The user creates a `Repository` object. If the `Repository` doesn't have a discovery schedule specified, our mutating admission controller will put a default one.
> 
>     1. Does `Repository` object mean `Registry` object here?

Yes, you're right. I'll fix that into  the original comment.

>     2. So we will need a mutating admission controller in sbombastic, correct? (when no discovery schedule specified, put a default discovery schedule and set suspend=true). Do we need to mutate

We don't need to have a mutating admission controller for that. We can leverage kube-builder `default` annotation to specify that value (see [here](https://book.kubebuilder.io/reference/markers/crd-validation.html)).
Also, I would not set `suspend` to be true by default, quite the opposite, it should be false by default to ensure the registry is regularly updated.

> Is my understanding correct : In k8s only the leader controller pod will be reconciling events while the other two will be on standby

Yes

> I thought the ticker loop runs in every controller but a controller creates instances of `RegistryDiscovery` only when it's the leader? (with the help of leaderelection.RunOrDie())

We have already gone through a careful investigation of how other projects deal with that aspect. That's how we ended up with the solution adopted by Kubewarden, the one I linked previously.
 
> > Finally, we could maybe take all the `RegistryDiscovery` jobs the user/system creates. Then use a Kubernetes Lease (one per registry ofc), to ensure only one job is processed at the same time.
> 
> Do you mean when a `RegistryDiscovery` job is running for registry-A, no `RegistryDiscovery` job for all registries can run at the same time? The "two `RegistryDiscovery` jobs cannot run at the same time" limitation is per-registry or for-all-registries ?

The lease should be taken on a per-registry basis. Only one discovery job per registry can be running at the same time.

Of all the approaches I proposed, this is the one I personally prefer. However, we still have to dig a bit deeper into NATs' documentation to see if there's any facility we can use to achieve the same result. Right now I'm a bit under the water and I cannot search for it, do you want to take a stab at that?

pohanhuangtw:
It seems that the current proposal doesn't address the re-scan mechanism. 
Should we consider adding a re-scan process? Some aspects we might want to cover include:

>External Changes:
  - Trivy DB update (every 6 hours) -> generate new report?
  - scanner version change, trigger the generate sbom -> generate new report?

>Avoiding Redundant Scans:
  - How do we determine if an image or its SBOM has been scanned before?
    - For example, if a registry scan fails midway, should we maintain a cache of successfully scanned images, or should we always query the current DB state to verify? or we needs to maintain a list to verify?

pohanhuangtw:
## Feedback for the RFC
### On the Discovery Job CR

>Question

Should the Discovery Job CR be removed after it completes?

>Suggestion

I think it should be removed, and here’s why:

1. Owner Reference Trigger: Currently, deleting a Discovery Job CR triggers its Owner Reference, which is tied to the registry. Keeping the CR might complicate or unnecessarily retain this dependency.
2. Resource Overhead: The system must repeatedly create new CRs for each scan. If users need to schedule daily scans across multiple registries, this approach generates excessive CRs, which feels wasteful.
3. Improvement Idea:
Consider a lighter-weight mechanism (e.g., direct scheduler management) to reduce the overhead of creating and deleting CRs.

### On Message Handling with NATS

> Question

Does every message get pushed to NATS by the leader?

>Suggestion

If so, this could simplify triggering rescans based on database version changes or scheduled scans, resolving these cases efficiently.

>Rationale

Centralizing message pushes through the leader could streamline rescan logic and ensure consistency.
@flavio does this resolve the open question you have?

### On the NATS Server

> Question

The NATS server is currently embedded in the controller. Should we separate it?

> Concern

If the controller Pod goes down, does this mean the data in NATS is lost?

> Thoughts

Embedding NATS in the controller ties its lifecycle to the Pod, risking data loss on crashes (e.g., during leader election transitions).

Does separating it out consider as next improvement?

flavio:
> Should the Discovery Job CR be removed after it completes?

I think we can leave some of them around. In my proposal the Registry spec has two fields that are used to specify how many successful and failed runs have to be kept around.

>     1. Owner Reference Trigger: Currently, deleting a Discovery Job CR triggers its Owner Reference, which is tied to the registry. Keeping the CR might complicate or unnecessarily retain this dependency.

That's wrong. If you delete a DiscoveryJob that doesn't lead to a reconciliation of its parent (the Registry), unless we instruct the Registry controller to do that (which we should not do).

>     2. Resource Overhead: The system must repeatedly create new CRs for each scan. If users need to schedule daily scans across multiple registries, this approach generates excessive CRs, which feels wasteful.

I don't think this should be a concern. The `DiscoveryJob` resource is small, we have a garbage collection mechanism in place (see the proposal). Plus I don't expect a user to define thousands of Registry objects/

> Does every message get pushed to NATS by the leader?

Yes. The jobs are added to the NATS queue by controllers, the controllers are reconciling events only on the leader. 

> Centralizing message pushes through the leader could streamline rescan logic and ensure consistency. @flavio does this resolve the open question you have?

No it doesn't. There are still cases (mentioned above) that are not solved by that.

> The NATS server is currently embedded in the controller. Should we separate it?

No, this simplifies the deployment for the end user. It's also a fully supported deployment mechanism of NATS.

In the future we might want to allow the user to specify an external NATS instance managed by them, just in case they already have a NATS cluster already running.

> If the controller Pod goes down, does this mean the data in NATS is lost?

Yes, but the data in NATS is ephemeral for us. We can recreate all of that. Moreover, this is a problem only if the user is running with 1 instance of the controller.

> Embedding NATS in the controller ties its lifecycle to the Pod, risking data loss on crashes (e.g., during leader election transitions).

I don't see the connection between the leader election transition and NATS. These are two different topics.

However, the risk doesn't exist when running with replica size major than 1. Unless all/a majority of them goes down at the same time. If that happens, that's not a big deal, since the queue will be re-populated accordingly by the scheduled tasks.

BTW, our controller is deployed as a StatefulSet because we have NATS running embedded inside of them.

pohanhuangtw:
Hi @flavio ,
Thanks for confirming:

1. The DiscoveryJob removal and resource concerns are not an issue.
2.  Embedding NATS within the controller Pod simplifies deployment, and we agree it's a reasonable default.

However, I’d like to raise a few points regarding the current NATS deployment and how it behaves under a high-availability (HA) setup.

# Data loss
## Scenario

1. The sbombastic controller is deployed in HA mode (e.g., StatefulSet with multiple replicas).
2. Only the leader’s NATS server (say, nat-server-1) publishes messages to the queue.
3. The controller Pod hosting nat-server-1 goes down unexpectedly.

If we don’t explicitly configure a Persistent Volume Claim (PVC) for nat-server-1’s data (currently stored in /tmp/nats/jetstream with nats.FileStorage), all in-flight messages—queued but not yet consumed—will be lost when the Pod restarts. This could disrupt workflows, especially if the leader crashes right after publishing a job.

## Proposed solution: JetStream Clustering

- To address this, I suggest setting up a JetStream cluster across the controller Pods, while still having the leader publish messages.jetstream cluster, ensuring the durability.
- Implementation: Embed a NATS JetStream cluster in the StatefulSet, with each Pod running a NATS server. The leader publishes to the cluster’s stream, which replicates data to all nodes.

Examples:
[Marmot’s embedded NATS clustering](https://github.com/maxpert/marmot/blob/fe252af4a57e472e21455acdbbaec0dcfd12221d/stream/embedded_nats.go#L40)
[AnyCable’s embedded NATS setup](https://github.com/anycable/anycable/blob/65af88ae6ddc16688a48ba507b5c9392b42b6272/enats/enats.go#L94)
#### Pros:

1. Data Resilience: Even if the leader’s Pod crashes, the JetStream cluster retains the queue data across other nodes, preventing loss.
2. Caching Potential: We could set MaxAge: 6 * time.Hour on the stream, using it as a distributed cache to skip redundant scans within that window.

#### Potential Issues
- Single Replica Case: If the user sets replicas: 1 in the StatefulSet, clustering won’t work—JetStream requires at least 3 nodes for Raft consensus, just fall back to single node mode?
  - Warn users in docs to use replicas >= 3 for HA.

### Questions for You
1. Do you think JetStream clustering resolves the data loss edge case you mentioned (e.g., leader crash post-publish)? Or is there another scenario I’m missing?
2. For the caching idea (6-hour retention for fitting trivy db update frequency), would that fit our use case, or do you see downsides to storing jobs that long?
3. How should we handle the single-replica fallback—strict enforcement of HA, or just a warning and let it run?

pohanhuangtw:
Hi @flavio ,
Thanks for confirming:

1. The DiscoveryJob removal and resource concerns are not an issue.
2.  Embedding NATS within the controller Pod simplifies deployment, and we agree it's a reasonable default.

However, I’d like to raise a few points regarding the current NATS deployment and how it behaves under a high-availability (HA) setup, [example commit](https://github.com/pohanhuangtw/sbombastic/commit/31935869d12b55cad5b98815a030c0d06a9bd428).



# Data loss
## Env setup

1. 3 controller, has pvc for each of the pod, and only leader can publish [create catalog
](https://github.com/pohanhuangtw/sbombastic/commit/31935869d12b55cad5b98815a030c0d06a9bd428#diff-f010c36608d27901df4f99a379382cd32fb74b4632c2a7cb21a4f2ec835b76e9R45-R51)
2. 3 worker
3. 1 storage (for easy reference)

## Data inconsistance Scenario
### Result
1. the reconciler only works in lead, does this expected ? @fabriziosestito 
2. every worker can receive the message through nats
3. Data lost when the controller pod is down
Screenshot 2025-04-02 at 5 45 09 PM

### How to reproduce
1. The sbombastic controller is deployed in HA mode (e.g., StatefulSet with multiple replicas).
  -  Controller N has nat-server-N as nat server
4. Controller 1 has NATS instance 1 (acting as the leader) responsible for publishes messages to the queue.
5. The controller Pod hosting nat-server-1 goes down unexpectedly.
6. Now controller 2 is the leader.
7. Since only the leader can publish message to nats, the controller 1 back to life with pvc backup
8. kill controller 2, so on so forth


## Proposed solution: JetStream Clustering

- To address this, I suggest setting up a JetStream cluster across the controller Pods, while still having the leader publish messages.jetstream cluster, ensuring the durability.
- Implementation: Embed a NATS JetStream cluster in the StatefulSet, with each Pod running a NATS server. The leader publishes to the cluster’s stream, which replicates data to all nodes.

Examples:
[Marmot’s embedded NATS clustering](https://github.com/maxpert/marmot/blob/fe252af4a57e472e21455acdbbaec0dcfd12221d/stream/embedded_nats.go#L40)
[AnyCable’s embedded NATS setup](https://github.com/anycable/anycable/blob/65af88ae6ddc16688a48ba507b5c9392b42b6272/enats/enats.go#L94)
#### Pros:

1. Data Resilience: Even if the leader’s Pod crashes, the JetStream cluster retains the queue data across other nodes, preventing loss.
4. Caching Potential: We could set MaxAge: 6 * time.Hour on the stream, using it as a distributed cache to skip redundant scans within that window.

#### Potential Issues
- Single Replica Case: If the user sets replicas: 1 in the StatefulSet, clustering won’t work—JetStream requires at least 3 nodes for Raft consensus, just fall back to single node mode, and use pvc to store the nats data.
  - Warn users in docs to use replicas >= 3 for HA.

### Questions for You
1. Do you think JetStream clustering resolves the data loss edge case you mentioned (e.g., leader crash post-publish)? Or is there another scenario I’m missing?
2. . How should we handle the single-replica fallback—strict enforcement of HA, or just a warning and let it run?

williamlin-suse:
> 
> > Does every message get pushed to NATS by the leader?
> 
> Yes. The jobs are added to the NATS queue by controllers, the controllers are reconciling events only on the leader.
For the non-active controllers, if they don't reconcile Registry CRs when they are not active, we need to figure out how to make them know the Registry CRs when one of them takes the leader role. 
Or a controller needs to know when to list all Registry CRs (when it becomes the leader)

williamlin-suse:
> > For the non-active controllers, if they don't reconcile Registry CRs when they are not active, we need to figure out how to make them know the Registry CRs when one of them takes the leader role.
> > Or a controller needs to know when to list all Registry CRs (when it becomes the leader)
> 
> If I understand what you meant, there's nothing we have to do. When a controller grabs the becomes a leader the controllers will start to reconcile resources. The controller libraries will automatically list all the resources being monitored and will trigger a reconciliation loop per each resource found.
> 
> BTW, the "full reconciliation" is also done every X hours (I don't remember the default value right now, but there's a parameter to tune that) to mitigate the accidental loss of some event.

Good to know that.
So it means when a non-leader controller becomes a leader, all Registry/Image/SBOM/RegistryDiscovery CRs will be re-reconciled, correct? (meaning we have the chance to see what might be missed in the processing)

fabriziosestito:
> Questions for You
> 
>     Do you think JetStream clustering resolves the data loss edge case you mentioned (e.g., leader crash post-publish)? Or is there another scenario I’m missing?


@pohanhuangtw the current POC doesn’t cover NATS clustering.
That said, clustering is required for high availability, or more precisely, for a production-ready setup.
We’re already running a JetStream cluster and using a StatefulSet for the controller.

Adding support should be straightforward: we can introduce CLI flags to accept peer addresses (which are deterministic with StatefulSets), then construct the ClusteringOptions when embedding the NATS server here:
https://github.com/rancher-sandbox/sbombastic/blob/3ad23aa1a04e8b32568b2ceba91ff60377dd04b9/internal/messaging/nats.go#L14

ClusteringOptions:
https://pkg.go.dev/github.com/nats-io/nats-streaming-server/server#ClusteringOptions

It would follow a pattern similar to the examples you linked above.

>   How should we handle the single-replica fallback—strict enforcement of HA, or just a warning and let it run?

We should just document that for HA, the recommended replica count is 3 or 5, and set 3 as the default.

pohanhuangtw:
Hi @fabriziosestito ,
let move the discussion to [Feature Request: Implement Embedded NATS JetStream Clustering for HA and Data Resilience](https://github.com/rancher-sandbox/sbombastic/issues/170)

pohanhuangtw:
Hi team,

Since this discussion has diverged from the original topic, I’d like to summarize the key points so we can stay aligned and avoid an open-ended thread.

## TODOs

- [ ] **Introduce a new `RegistryDiscovery` CRD**  
  Used to manually trigger scan jobs via:  
  - [ ] Creating a new CR instance  
  - [ ] Clicking a "Discovery" action (UI/integration TBD)

- [ ] **Extend the `Registry` CRD with scheduling capabilities**  
  Inspired by how [[Kubewarden’s cert controller](https://github.com/kubewarden/kubewarden-controller/blob/ba6abb0742e68c3d26ab523d05b7ad009c48ca8b/internal/controller/cert_controller.go#L34)](https://github.com/kubewarden/kubewarden-controller/blob/ba6abb0742e68c3d26ab523d05b7ad009c48ca8b/internal/controller/cert_controller.go#L34) handles rotation.  
  - Solves the need for **periodic scanning**.

- [ ] **Move toward an HA (High Availability) setup**  
  - Enables per-registry scanning with better fault tolerance  
  - Since only the leader reconciles, JetStream would only receive events from the leader

## Related Issue

- [Feature Request: Implement Embedded NATS JetStream Clustering for HA and Data Resilience](https://github.com/rancher-sandbox/sbombastic/issues/170)](https://github.com/rancher-sandbox/sbombastic/issues/170)

Please feel free to share your thoughts or concerns!

PR #141, opened 2025-03-24
fix: add continue to avoid blocking the image generate.

pohanhuangtw:
## Description
Fix https://github.com/rancher-sandbox/sbombastic/issues/121

## Test
Inside of the development environment, with tilt up, create the following registry definition:
```yaml
apiVersion: sbombastic.rancher.io/v1alpha1
kind: Registry
metadata:
  name: kw-controller
  namespace: default
spec:
  uri: ghcr.io
  repositories:
    - kubewarden/kubewarden-controller   
```
The scan succeed, Image resources are created.

### DIFF
--- a/internal/handlers/create_catalog.go
+++ b/internal/handlers/create_catalog.go
@@ -107,7 +107,7 @@ func (h *CreateCatalogHandler) Handle(message messaging.Message) error {
 
 		images, err := h.refToImages(registryClient, ref, registry)
 		if err != nil {
-			h.logger.Error("Error cannot get images for ref", "ref", ref, "error", err)
+			h.logger.Error(fmt.Sprintf("cannot get images for %s: %s", ref, err.Error()))
 			// Avoid blocking other images to be cataloged
 			continue
 		}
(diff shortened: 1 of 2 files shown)


internal/handlers/create_catalog.go
```diff
@@ -107,7 +107,9 @@ func (h *CreateCatalogHandler) Handle(message messaging.Message) error {
 
 		images, err := h.refToImages(registryClient, ref, registry)
 		if err != nil {
-			return fmt.Errorf("cannot get images for %s: %w", ref, err)
+			h.logger.Error("Error cannot get images for ref", "ref", ref, "error", err)
```
flavio, line 110: nit, "Error" would already be part of the produced output

```suggestion
			h.logger.Error("cannot get images for ref", "ref", ref, "error", err)
```

flavio (changes requested):
Great, this is exactly what I was looking for.

Can you please expand the unit tests with a test case that covers this scenario?

internal/handlers/create_catalog.go
pohanhuangtw: Hi @flavio , 
After dig into the unit-test, I still not sure where to put the cases fir this PR
Any good suggestions?

flavio: inside of `create_catalog_test.go`. Take inspiration from something like `TestCreateCatalogHandler_Handle`.

I would create a new test, mock a repository with a container image and an OCI artifact inside of it

pohanhuangtw: Hi @flavio ,
I noticed that our current code may have issues when processing multiple platforms in a single image. For example, if we encounter a list like [arm64, amd, unknown (which causes a failure), other-platform], our code stops processing after the failure ("unknown"), and we miss generating images for subsequent platforms.

Should we adjust the logic to handle such errors gracefully, so that a failure on one platform doesn’t block processing of the others?

flavio: > Should we adjust the logic to handle such errors gracefully, so that a failure on one platform doesn’t block processing of the others?

For sure. I leave up to you to decided whether we should fix that with another PR. If you want to fix that with another PR, please write a new issue with a detailed explanation. When fixing the issue also provide the unit tests for that

pohanhuangtw: I will also fix this within the same pr with next commit

internal/handlers/create_catalog_test.go
```diff
@@ -101,7 +114,7 @@ func TestCreateCatalogHandler_Handle(t *testing.T) {
 
 	mockRegistryClient.On("GetImageDetails", image, &platformLinuxAmd64).Return(imageDetailsLinuxAmd64, nil)
 	mockRegistryClient.On("GetImageDetails", image, &platformLinuxArm64).Return(imageDetailsLinuxArm64, nil)
-
+	mockRegistryClient.On("GetImageDetails", image, mock.Anything).Return(registry.ImageDetails{}, fmt.Errorf("cannot get platform for %s", image))
```
flavio, line 117: I don't like using `mock.Anything`, I would prefer to be more specific.

Doing something like that works:

```golang
var nilPlatform *cranev1.Platform
mockRegistryClient.On("GetImageDetails", image, nilPlatform).Return(registry.ImageDetails{}, fmt.Errorf("cannot get platform for %s", image))
```

flavio (changes requested):

internal/handlers/create_catalog_test.go
pohanhuangtw: Thanks for the good suggestion.

PR #149, opened 2025-03-26
feat(test): create helm unit test

pohanhuangtw:
## Description

- Fix https://github.com/rancher-sandbox/sbombastic/issues/134
  - Ensure the helm test to be able to run in local / ci
  - Add test folder for each component 

## Test
Run `make helm-unittest `in the working dir.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,6 +18,11 @@ jobs:
         with:
           go-version: "1.24.1"
       - run: make test
+      - name: Install helm
+        uses: azure/setup-helm@b9e51907a09c216f16ebe8536097933489208112 # v4.3.0
+      - name: Install Helm-unittest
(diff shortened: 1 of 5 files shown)


pohanhuangtw:
Hi @flavio,
This is expected to fail because the SBOM is not yet updated.
We need to merge [PR #150](https://github.com/rancher-sandbox/sbombastic/pull/150) first — once that’s in, this should work as intended.

flavio:
This looks good to me, but I would prefer to have that as a separate job instead of collapsing that into the `Test` one (which is more about go code).

@fabriziosestito what do you think?

pohanhuangtw:
Hi @flavio ,
Does it mean we want a separate workflow or another step within the current workflow?

fabriziosestito:
@pohanhuangtw I would say another job within the same workflow so that they can run in parallel

flavio (approved):
nice, thanks!

PR #156, opened 2025-03-31
feat: Use PVC to handle the sqlite database

pohanhuangtw:
## Description

1. Fix https://github.com/orgs/rancher-sandbox/projects/6/views/1?pane=issue&itemId=102825107&issue=rancher-sandbox%7Csbombastic%7C128
2. Support PVC to handle the sqlite database
  - User needs to have it own storage class
  - If running from scratch with default settings, you need to create a PersistentVolume (PV) matching the PVC specs for SQLite data persistence. 
  - Add corresponding unit-test
3. Improve Readme in both main readme and helm chart readme

## Test
1. tilt up
5. deploy and ensure the result in the local file, reference to examples/pv.yaml
```yaml
apiVersion: sbombastic.rancher.io/v1alpha1
kind: Registry
metadata:
  name: kw-controller
  namespace: default
spec:
  uri: ghcr.io
  repositories:
    - kubewarden/kubewarden-controller
```


helm/templates/storage/README.md
```diff
@@ -0,0 +1,12 @@
+# SBOMbastic Storage
+
+The `storage` Helm chart installs the SBOMbastic storage deployment, which should be installed alongside the SBOMbastic controller and worker components.
+
+The storage component uses SQLite as its database backend. To ensure data persistence, it requires a PersistentVolumeClaim (PVC) backed by a PersistentVolume (PV). Users need to either:
+
+1. Have a StorageClass configured that can dynamically provision PVs, or
+2. Manually create a PV that matches the PVC requirements
```
flavio, line 8: Usually charts let the user create the PVC on their own, with their own criteria, and the provide that to the helm chart as a value.

I'll provide more details later

helm/templates/storage/README.md
```diff
@@ -0,0 +1,12 @@
...
+
+## Prerequisites
+
+When using the default configuration, you must ensure a suitable PersistentVolume exists before installing the chart. See `examples/pv.yaml` for a reference configuration.
```
flavio, line 12: This is basically a repetition of lines 5-8. Just move these lines done there, replacing this one.

helm/templates/storage/deployment.yaml
```diff
@@ -23,3 +23,10 @@ spec:
         - name: storage
           image: {{ .Values.storage.image.repository }}:{{ .Values.storage.image.tag }}
           imagePullPolicy: {{ .Values.storage.image.pullPolicy }}
+          volumeMounts:
+            - name: sqlite-pvc
+              mountPath: /data/sqlite/
+      volumes:
+        - name: sqlite-pvc
+          persistentVolumeClaim:
+            claimName: {{ .Values.storage.pvc.name }}
```
flavio, line 32: To be handled differently, see for example what is being done [here](https://github.com/nextcloud/helm/blob/00d8a86ba8bab3f10b7180946dfe37615a071f4c/charts/nextcloud/templates/deployment.yaml#L350-L354).

Something along these lines:
```yaml
  - name: data-sqlite
    persistentVolumeClaim:
            claimName: {{ if .Values.persistence.storageData.existingClaim }}{{ .Values.persistence.storageData.existingClaim }}{{- else }}{{ template "storage.fullname" . }}-sqlite-data{{- end }}
```

Double check what the `template staroge.fullname` does.

helm/templates/storage/pvc.yaml
flavio, line 1: Should be created only when the user does not provide an already existing claim to reuse.

Something similar to [that](https://github.com/nextcloud/helm/blob/main/charts/nextcloud/templates/nextcloud-data-pvc.yaml)

helm/templates/storage/storage_class.yaml
flavio, line 1: Remove, it's not our duty to provide a storage class.

helm/values.yaml
```diff
@@ -15,6 +15,16 @@ storage:
     tag: v0.1.0-alpha1
     pullPolicy: IfNotPresent
   replicas: 3
+  pvc:
+    name: sqlite-pvc
+    requestStorage: 1Gi
+    accessModes:
+      - ReadWriteMany
```
flavio, line 22: `pcv` is cryptic as a name. 

Take inspiration from what other charts do, for example [here](https://github.com/nextcloud/helm/blob/00d8a86ba8bab3f10b7180946dfe37615a071f4c/charts/nextcloud/values.yaml#L597-L613)

pohanhuangtw:
Hi @flavio,

Thanks for the advice. In the last two commits, I've removed the unnecessary files and fixed the related tests.

Regarding the storage component, I propose we deploy it as a StatefulSet instead of a Deployment. Since the component is currently using SQLite, deploying it as a StatefulSet ensures that Pods are created and deleted in a predetermined order, and during rollouts only one Pod is active at any given time. 
This setup minimizes the risk of having multiple Pods simultaneously accessing the same SQLite database and, I believe, is a safer approach than using a Deployment.

Do you think I should address this change in this PR?

helm/templates/storage/pvc.yaml
```diff
@@ -1,11 +1,28 @@
...
     requests:
-      storage: {{ .Values.storage.pvc.requestStorage }}
\ No newline at end of file
+      storage: {{ .Values.persistence.storageData.size | quote }}
+  {{- with .Values.persistence.storageData.storageClass }}
+  {{- if (eq "-" .) }}
+  storageClassName: standard
+  {{- else }}
+  storageClassName: "{{ . }}"
+  {{- end }}
```
flavio, line 26: the `standard` storage class is not granted. Many k8s distro have it, but it's not guaranteed to be present.
Using an empty name is safer, since it will fall back to use whatever storage class has been marked as the default one by the admin.

Finally, I'm not super convinced about this handling of the `-` symbol. I would just propagate whatever the user put into the `values`. However, if you want to keep the `-` thing around, at least document that inside of the values.yaml

flavio (changes requested):

helm/templates/storage/README.md
```diff
@@ -2,9 +2,13 @@
...
-To ensure data persistence, it requires a PersistentVolumeClaim (PVC) backed by a PersistentVolume (PV). Users need to either:
+To ensure data persistence, the storage component requires a PersistentVolumeClaim (PVC). You can provide your own PVC to control how and where data is stored. This is the recommended approach and aligns with common Helm chart practices.
 
-1. Have a StorageClass configured that can dynamically provision PVs, or
-2. Manually create a PV that matches the PVC requirements
\ No newline at end of file
+There are two ways to satisfy this requirement:
+
+1. **Recommended:** Provide a pre-created PVC and reference it in your Helm values using `persistence.storageData.existingClaim`.
+2. If no PVC is provided, and your cluster supports dynamic provisioning via a StorageClass, a new PVC and corresponding PV will be created automatically.
```
flavio, line 12: Why the 1st solution is the recommended one? I personally think the 2nd one is the simplest one, since it's all "self-service".


```suggestion
1. Provide a pre-created PVC and reference it in your Helm values using `persistence.storageData.existingClaim`.
2. If no PVC is provided, and your cluster supports dynamic provisioning via a `StorageClass`, a new PVC and corresponding PV will be created automatically.
```

helm/templates/storage/README.md
```diff
@@ -2,9 +2,13 @@
...
+
+If dynamic provisioning is not available and you don't provide a PVC, you will need to manually create a PersistentVolume that matches the chart's default PVC spec.
```
flavio, line 14: Not really needed:

```suggestion
```

helm/templates/storage/README.md
```diff
@@ -2,9 +2,13 @@
 
 The `storage` Helm chart installs the SBOMbastic storage deployment, which should be installed alongside the SBOMbastic controller and worker components.
 
-The storage component uses SQLite as its database backend. **Note that SQLite is intended for development and testing purposes only, and should not be used in production environments.** 
+The storage component uses SQLite as its database backend. **Note that SQLite is intended for development and testing purposes only, and should not be used in production environments.**
 
-To ensure data persistence, it requires a PersistentVolumeClaim (PVC) backed by a PersistentVolume (PV). Users need to either:
+To ensure data persistence, the storage component requires a PersistentVolumeClaim (PVC). You can provide your own PVC to control how and where data is stored. This is the recommended approach and aligns with common Helm chart practices.
```
flavio, line 7: nit

```suggestion
To ensure data persistence, the storage component requires a PersistentVolumeClaim (PVC). You can provide your own PVC to control how and where data is stored.
```

helm/tests/storage/pvc_test.yaml
```diff
@@ -0,0 +1,73 @@
...
+    set:
+      persistence:
+        enabled: true
+        storageData:
+          enabled: true
+          storageClass: "-"
+    asserts:
+      - equal:
+          path: "spec.storageClassName"
+          value: "-"
```
flavio, line 40: The description is no longer relevant, you're no longer emitting the `standard` name. 

We should check the following things:
- the name of the storageClass provided by the user is propagated. Kinda what you're doing here, but I would use valid name like `fast`.
- ensure an empty storageClass name is still propagated as empty string

helm/values.yaml
```diff
@@ -14,11 +14,26 @@ storage:
     repository: ghcr.io/rancher-sandbox/sbombastic/storage
     tag: v0.1.0-alpha1
     pullPolicy: IfNotPresent
-  replicas: 3
+  replicas: 1
```
flavio, line 17: Please, file an issue about whether we should keep using `Deployment` to run the storage component, as opposed to `StatefulSet`. I left a comment about that above.

helm/tests/storage/pvc_test.yaml
```diff
@@ -0,0 +1,84 @@
...
+          path: "spec.storageClassName"
+          value: "sqlite-storageclass-test"
+      - equal:
+          path: "spec.accessModes"
+          value:
+          - ReadWriteMany
+      - equal:
+          path: "spec.resources.requests.storage"
+          value: "10Gi"
+  - it: "should use standard storage class when storageClass is set to any user-defined storage class"
```
flavio, line 30:
```suggestion
  - it: "should use the storage class defined by the user"
```

helm/tests/storage/pvc_test.yaml
```diff
@@ -0,0 +1,84 @@
...
+      persistence:
+        enabled: true
+        storageData:
+          enabled: true
+          storageClass: "fast-storageclass-test"
+    asserts:
+      - equal:
+          path: "spec.storageClassName"
+          value: "fast-storageclass-test"
+  - it: "should use allow usage of an empty string storage class to allow cluster's default one to be used"
```
flavio, line 41: sorry, I made a typo:

```suggestion
  - it: "should allow the usage of an empty string storage class to allow cluster's default one to be used"
``

flavio (dismissed):

flavio:
I've grouped the commits together to keep the history meaningful. Once everything is green I'll merge it

flavio (approved):

PR #158, opened 2025-04-01
feat: use pflag to parse the args from helm

pohanhuangtw:
## Description

- Fix https://github.com/rancher-sandbox/sbombastic/issues/29
- Replace flag with pflag to fix parsing issues in the storage component
- Add Helm template unit test for passing --log-level via args

## Test

1. tilt up
2. able to see the log level set from the value.yaml

### Potential improvement
- Move flag parsing to internal/cmdutil/flag.go and add unit tests


internal/cmdutil/log_test.go
flavio, line 1: It's nice to have a test for that, but this is really testing that slog `UnmarshalText` works as they advertise.

@fabriziosestito should we even drop this test?

internal/cmdutil/log.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"fmt"
+	"log/slog"
+)
+
+func ParseLogLevel(s string) (slog.Level, error) {
+	var level slog.Level
+	if err := level.UnmarshalText([]byte(s)); err != nil {
+		return level, fmt.Errorf("unable to parse log level: %w", err)
```
flavio, line 11: all the code that invokes this helper is then returning the error message doing something like that:

```golang
slog.Error("unable to parse log level", "error", err)
os.Exit(1)
```

I see two problems:

- the final output is going to contain the same text repeated twice ("unable to parse log level")
- there's some boilerplate: `if parseloglevel returns error; print error && exit ; endif`


I prose we change this code to be something like:

```golang
// ParseLogLevelOrFail attempts to convert the given log
// level to a `slog.Level`. The function will exit the program
// with an error if the conversion fails
func ParseLogLevelOrFail(s string) slog.Level {
	var level slog.Level
	if err := level.UnmarshalText([]byte(s)); err != nil {
		slog.Error("unable to parse log level: %w", err)
                os.Exit(1)
        }
        return level
}                
```

flavio (changes requested):

internal/cmdutil/log_test.go
fabriziosestito: i think we can drop it

internal/cmdutil/log_test.go
```diff
@@ -0,0 +1,80 @@
+package cmdutil
+
+import (
+	"errors"
+	"fmt"
+	"log/slog"
+	"testing"
+
+	"gotest.tools/v3/assert"
```
fabriziosestito, line 9: we are using testify already, let's not add other libraries in the future

internal/cmdutil/log.go
fabriziosestito: ​@flavio I think the current implementation is fine. Invoking os.Exit outside the main function can complicate maintenance and testing. For instance calling `os.Exit` directly within functions other than main does not execute `defer func` blocks. 
Also using a global logger here seems not ideal, as we have a linter exception only for the main  function.

Instead, we can modify the caller's log message to output only the error, or use a generic message like "error initializing the logger: unable to parse log level."

cmd/controller/main.go
```diff
@@ -40,43 +39,60 @@ import (
...
+	EnableLeaderElection bool
+	SecureMetrics        bool
+	EnableHTTP2          bool
+	LogLevel             string
+}
 
-	flag.StringVar(&metricsAddr, "metrics-bind-address", "0", "The address the metrics endpoint binds to. "+
+func parseFlags() Config {
+	var cfg Config
+	pflag.StringVar(&cfg.MetricsAddr, "metrics-bind-address", "0", "The address the metrics endpoint binds to. "+
```
fabriziosestito, line 60: Why are we switching to pflags? What are the advantages?

internal/cmdutil/log.go
flavio: good points @fabriziosestito :+1:

cmd/controller/main.go
pohanhuangtw: I think we should make the flag usage consistent with the other components, since the storage component is already using pflag.

flavio: I was surprised too about the switch. I think we could switch everything to [`flag`](https://pkg.go.dev/flag) to reduce the dependency tree.

I think storage might have been using pflag because it's a project that was scaffolded from a template. We have nothing special that requires it AFAIK

flavio: Ok, but now the controller has been switched to `flags` and all the tests are green. I'm confused :thinking: :smile:

go.mod
```diff
@@ -22,6 +22,7 @@ require (
 	github.com/spdx/tools-golang v0.5.5
 	github.com/spf13/cobra v1.9.1
 	github.com/stretchr/testify v1.10.0
+	gotest.tools/v3 v3.4.0
```
flavio, line 25: This is not needed, you forgot to run `go mod tidy`

cmd/storage/main.go
```diff
@@ -23,17 +23,27 @@ import (
...
 	"k8s.io/component-base/cli"
 
 	_ "modernc.org/sqlite"
 )
 
 func main() {
-	// TODO: add CLI flags
+	var logLevel string
+	pflag.StringVar(&logLevel, "log-level", slog.LevelDebug.String(), "Log level")
+	pflag.Parse()
```
flavio, line 38: This can be migrated to `flag`

cmd/storage/main.go
```diff
@@ -23,17 +23,27 @@ import (
...
+
+	slogLevel, err := cmdutil.ParseLogLevel(logLevel)
+	if err != nil {
+		slog.Error("unable to parse log level", "error", err) //nolint:sloglint // Use the global logger since the logger is not yet initialized
```
flavio, line 42: Update the error message as discussed before

internal/cmdutil/log.go
flavio: @pohanhuangtw you have still to update the log messages of the consumers of this function. See the comment left above

flavio (changes requested):

pohanhuangtw:
Hi @flavio ,
Are you suggesting we log in the function not the caller?

flavio:
> Hi @flavio , Are you suggesting we log in the function not the caller?

No, I'm referring to what I wrote before: the final output is going to contain the same text repeated twice ("unable to parse log level")

cmd/controller/main.go
pohanhuangtw: I’ve switched the controller/worker back to use flags, but there's currently no test for it.

cmd/storage/main.go
flavio: I see

flavio: I've looked deeper into this topic. It turns out the kubernetes libraries we're using are using cobra and are configuring a lot of flags against it, plus - that's the crucial part - they are making use of `klog`.

I think that handling our log level in this way is like fighting against the libraries we're using. For example, look at what happens when we run the following command:

```console
go run cmd/storage/main.go --help
Usage of /tmp/go-build4001753831/b001/exe/main:
      --log-level string   Log level (default "DEBUG")
pflag: help requested
exit status 2
```

For the sake of this PR, I would not do any change inside of storage. I would just create a GitHub issue to keep track of this technical debt. Meaning, we have to decide how to handle logging inside of storage. I see two options:

- surrender and use the kubernetes approach. Which means, rely on klog and on the log levels of it (the ones expressed as numbers via the `-v` flag that is automatically added to our cli program)
- attempt to change things by doing some heavy refactoring

helm/templates/controller/statefulset.yaml
```diff
@@ -25,6 +25,9 @@ spec:
           args:
             - --leader-elect
             - --health-probe-bind-address=:8081
+            {{- if .Values.controller.logLevel }}
+            - --log-level={{ .Values.controller.logLevel }}
```
flavio, line 29: These are now `-`, not `--`

```console
go run cmd/controller/main.go --help
Usage of /home/flavio/.cache/go-build/5e/5ec0c8754514ca72f73f542eab3aedd12a58f5e37e31d91f2312429d09d39f04-d/main:
  -enable-http2
        If set, HTTP/2 will be enabled for the metrics and webhook servers
  -health-probe-bind-address string
        The address the probe endpoint binds to. (default ":8081")
  -kubeconfig string
        Paths to a kubeconfig. Only required if out-of-cluster.
  -leader-elect
        Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.
  -log-level string
        Log level (default "INFO")
  -metrics-bind-address string
        The address the metrics endpoint binds to. Use :8443 for HTTPS or :8080 for HTTP, or leave as 0 to disable the metrics service. (default "0")
  -metrics-secure
        If set, the metrics endpoint is served securely via HTTPS. Use --metrics-secure=false to use HTTP instead. (default true)
```

helm/templates/worker/deployment.yaml
```diff
@@ -24,8 +24,11 @@ spec:
           image: {{ .Values.worker.image.repository }}:{{ .Values.worker.image.tag }}
           imagePullPolicy: {{ .Values.worker.image.pullPolicy }}
           args:
-            - -nats-url
+            - --nats-url
             - {{ .Release.Name }}-controller-nats.{{ .Release.Namespace }}.svc.cluster.local:4222
+            {{- if .Values.worker.logLevel }}
+            - --log-level={{ .Values.worker.logLevel }}
+            {{- end }}
```
flavio, line 31: These are now `-`, not `--`:

```console
 go run cmd/worker/main.go --help
Usage of /home/flavio/.cache/go-build/d1/d1e78061a54e8eb552aa999aff9d4fdfba8cd47c036c9005054423bc477484fc-d/main:
  -kubeconfig string
        Paths to a kubeconfig. Only required if out-of-cluster.
  -log-level string
        Log level (default "INFO")
  -nats-url string
        The URL of the NATS server (default "localhost:4222")
  -run-dir string
        Directory to store temporary files (default "/var/run/worker")
```

cmd/storage/main.go
pohanhuangtw: My opinion is to use klog instead of performing a heavy refactoring, because:

- Deep Integration with Kubernetes API:

We are highly integrated with the Kubernetes API, which relies on klog for logging. It makes sense to adopt klog so that our logging behavior aligns with the wider Kubernetes ecosystem.

- Effort and Priority Considerations:

Undertaking a major refactoring to replace our current logging approach would require significant effort, and given our current priorities, it isn’t the best use of our resources right now.

Overall, moving to klog allows us to streamline our logging configuration, avoid potential conflicts, and leverage the standard practices of the Kubernetes environment without the overhead of a major refactoring.

what do you think? Should we do this in the PR?

flavio: I would klog inside of storage. This seems a different topic compared to the current PR (using pflags to parse args inside of helm charts). File another issue please. Let's try to get this PR closed

pohanhuangtw: Hi @flavio 
issue https://github.com/rancher-sandbox/sbombastic/issues/167 has created.

flavio: As I wrote above,

> For the sake of this PR, I would not do any change inside of storage.

Just undo the changes done to this file. Then we're good to go

flavio: Yes, because the current code is just breaking the program if you run with `--help` (see comment from above)

pohanhuangtw: Just make the change, please help review! Thanks

flavio (approved):

pohanhuangtw:
Hi @fabriziosestito ,
Would you mind take a look at this PR?

pohanhuangtw:
Hi @williamlin-suse  ,
Would you mind take a look at this PR? Thanks!

flavio (approved):

fabriziosestito (approved):

Issue #173, opened 2025-04-11
Unable to run tilt up


pohanhuangtw:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

When running the tilt up, keeps showing error

```
Loading Tiltfile at: /home/ubuntu/suse/sbombastic/Tiltfile
Running: helm template sbombastic /home/ubuntu/suse/sbombastic/helm --namespace sbombastic --include-crds --set controller.image.repository=hhcs9527/controller --set storage.image.repository=hhcs9527/storage --set worker.image.repository=hhcs9527/worker --set controller.replicas=1 --set storage.replicas=1 --set worker.replicas=1
Traceback (most recent call last):
  /home/ubuntu/suse/sbombastic/Tiltfile:28:12: in <toplevel>
Error in helm: command "helm template sbombastic /home/ubuntu/suse/sbombastic/helm --namespace sbombastic --include-crds --set controller.image.repository=hhcs9527/controller --set storage.image.repository=hhcs9527/storage --set worker.image.repository=hhcs9527/worker --set controller.replicas=1 --set storage.replicas=1 --set worker.replicas=1" failed.
error: exit status 1
stdout:

stderr:
Error: YAML parse error on sbombastic/templates/storage/README.md: error converting YAML to JSON: yaml: line 9: mapping values are not allowed in this context

Use --debug flag to render out invalid YAML

```

### Expected Behavior

it should generate the helm and let kubernetes apply the yamls.

### Steps To Reproduce

1. in main branch
2. run tilt up
3. get into the ui, then will see the issue

### Environment

```markdown
- OS:os
- Architecture:amd64
```

PR #174, opened 2025-04-11
fix: tilt up error

pohanhuangtw:
## Description

- Fixes [issue #173](https://github.com/rancher-sandbox/sbombastic/issues/173)
-  root caus
   - When running `tilt up`, Helm is first called via `helm template` to render the chart.
   - Including `*.md` files in the chart directory causes Helm to fail, since Markdown files are not valid according to Helm's templating rules.

## Test


helm/values.yaml
flavio, line 1: nitpicking: this is still being changed

flavio (changes requested):

flavio (approved):

PR #176, opened 2025-04-14
feat: add k8s recommended labels

pohanhuangtw:
## Description
- Fix https://github.com/rancher-sandbox/sbombastic/issues/27
- Add the labels in the image / SBOM / VulnerabilityRerport, because these are created by controller

## Test
- tilt up 
- create a mock data provided in examples
- ensure the image / SBOM / VulnerabilityRerport has label

### DIFF
--- a/helm/templates/_helpers.tpl
+++ b/helm/templates/_helpers.tpl
@@ -39,6 +39,7 @@ helm.sh/chart: {{ include "sbombastic.chart" . }}
 {{- if .Chart.AppVersion }}
 app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
 {{- end }}
+app.kubernetes.io/part-of: sbombastic
 app.kubernetes.io/managed-by: {{ .Release.Service }}
 {{- end }}
 
--- a/internal/handlers/generate_sbom.go
+++ b/internal/handlers/generate_sbom.go
@@ -120,6 +120,10 @@ func (h *GenerateSBOMHandler) Handle(message messaging.Message) error {
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      generateSBOMMessage.ImageName,
 			Namespace: generateSBOMMessage.ImageNamespace,
+			Labels: map[string]string{
+				LabelManagedByKey: LabelManagedByValue,
+				LabelPartOfKey:    LabelPartOfValue,
+			},
 		},
 		Spec: storagev1alpha1.SBOMSpec{
 			ImageMetadata: image.GetImageMetadata(),
--- a/internal/handlers/scan_sbom.go
+++ b/internal/handlers/scan_sbom.go
@@ -41,6 +41,7 @@ func NewScanSBOMHandler(
 	}
 }
 
+//nolint:funlen
 func (h *ScanSBOMHandler) Handle(message messaging.Message) error {
 	scanSBOMMessage, ok := message.(*messaging.ScanSBOM)
 	if !ok {
@@ -133,6 +134,11 @@ func (h *ScanSBOMHandler) Handle(message messaging.Message) error {
 	}
 
 	_, err = controllerutil.CreateOrUpdate(ctx, h.k8sClient, vulnerabilityReport, func() error {
+		vulnerabilityReport.Labels = map[string]string{
+			LabelManagedByKey: LabelManagedByValue,
+			LabelPartOfKey:    LabelPartOfValue,
+		}
+
 		vulnerabilityReport.Spec = storagev1alpha1.VulnerabilityReportSpec{
 			ImageMetadata: sbom.GetImageMetadata(),
 			SARIF:         runtime.RawExtension{Raw: reportBytes},
--- a/helm/templates/storage/pvc.yaml
+++ b/helm/templates/storage/pvc.yaml
@@ -5,6 +5,8 @@ apiVersion: v1
 metadata:
   name: {{ template "sbombastic.fullname" . }}-storage-data
   labels:
+  {{ include "sbombastic.labels" .| nindent 4 }}
+  app.kubernetes.io/component: storage
   {{- with .Values.persistence.storageData.labels }}
     {{- toYaml . | nindent 4 }}
   {{- end }}
--- a/internal/handlers/constants.go
+++ b/internal/handlers/constants.go
@@ -1,0 +1,8 @@
+package handlers
+
+const (
+	LabelManagedByKey   = "app.kubernetes.io/managed-by"
+	LabelManagedByValue = "sbombastic"
+	LabelPartOfKey      = "app.kubernetes.io/part-of"
+	LabelPartOfValue    = "sbombastic"
+)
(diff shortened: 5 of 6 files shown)


pohanhuangtw:
Hi @flavio, @fabriziosestito, @williamlin-suse,
Could you help review this PR? I believe it's fairly straightforward — it ensures the recommended ABs are added to our CR.
Thanks!

pohanhuangtw:
Thanks @holyspectral for the quick review.

- I update the way I built the map.
- improve lable update in the create-update only when the label is nil.

internal/handlers/scan_sbom.go
```diff
@@ -129,6 +137,9 @@ func (h *ScanSBOMHandler) Handle(message messaging.Message) error {
 	}
 
 	_, err = controllerutil.CreateOrUpdate(ctx, h.k8sClient, vulnerabilityReport, func() error {
+		if vulnerabilityReport.Labels == nil {
+			vulnerabilityReport.Labels = prepareVulnerabilityReportLabels()
+		}
```
holyspectral, line 142: You can implement the real `mutate()` logic like this, so if the labels are removed somehow, they will be added  back.

```suggestion
		if vulnerabilityReport.Labels == nil {
			vulnerabilityReport.Labels = prepareVulnerabilityReportLabels()
		} else {
		        vulnerabilityReport.Labels["app.kubernetes.io/managed-by"] = LabelValueManagedBy
		        vulnerabilityReport.Labels["app.kubernetes.io/component"] = LabelValueComponent
		        vulnerabilityReport.Labels["app.kubernetes.io/part-of"] = LabelValuePartOf
		}
```

This is probably not something that matters in your scenario, but it's still a good idea to understand how `CreateOrUpdate()` works and how to avoid pitfall like this.  (Think about if you have more than one writer to add different labels to the CR.  )

Also it would be nice to move these logic to something like `mutateVulnerabilityReportLabels()` function.

pohanhuangtw: You're right about this. Would it be correct to say that these labels shouldn't be changed by users?
I did consider this approach, but I was concerned that users might modify the labels manually — is that a valid concern?

pohanhuangtw: you are right about that, removing is something we should concerned about, let me add it, thanks.

pohanhuangtw:
Hi @holyspectral ,
Just update the with you suggestions, thanks

internal/handlers/generate_sbom.go
```diff
@@ -105,6 +113,7 @@ func (h *GenerateSBOMHandler) Handle(message messaging.Message) error {
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      generateSBOMMessage.ImageName,
 			Namespace: generateSBOMMessage.ImageNamespace,
+			Labels:    prepareSBOMLabels(),
```
flavio, line 116: Also in this case I don't see the value of having this helper function

internal/handlers/registry/mocks/client.go
flavio, line 1: This change is not related with the PR, if possible I would add it to another PR

internal/handlers/registry/mocks/image_index.go
flavio, line 1: This change is not related with the PR, if possible I would add it to another PR

internal/handlers/registry/mocks/layer.go
flavio, line 1: This change is not related with the PR, if possible I would add it to another PR

internal/handlers/scan_sbom.go
```diff
@@ -36,6 +36,14 @@ func NewScanSBOMHandler(k8sClient client.Client, scheme *runtime.Scheme, workDir
 	}
 }
 
+func prepareVulnerabilityReportLabels() map[string]string {
+	return map[string]string{
+		"app.kubernetes.io/managed-by": LabelValueManagedBy,
+		"app.kubernetes.io/component":  LabelValueComponent,
```
flavio, line 42: Same here, the "component" label should not belong to the CRDs

internal/messaging/mocks/publisher.go
flavio, line 1: This change is not related with the PR, if possible I would add it to another PR

internal/messaging/publisher.go
flavio, line 1: This change is not related with the PR, if possible I would add it to another PR

internal/handlers/scan_sbom.go
```diff
@@ -129,6 +137,13 @@ func (h *ScanSBOMHandler) Handle(message messaging.Message) error {
 	}
 
 	_, err = controllerutil.CreateOrUpdate(ctx, h.k8sClient, vulnerabilityReport, func() error {
+		if vulnerabilityReport.Labels == nil {
+			vulnerabilityReport.Labels = prepareVulnerabilityReportLabels()
+		} else {
+			vulnerabilityReport.Labels["app.kubernetes.io/managed-by"] = LabelValueManagedBy
+			vulnerabilityReport.Labels["app.kubernetes.io/component"] = LabelValueComponent
+			vulnerabilityReport.Labels["app.kubernetes.io/part-of"] = LabelValuePartOf
+		}
```
flavio, line 146: I've seen the comment of @holyspectral about the risk of overwriting use labels. In this case however, we are the sole owners of the `VulnerabilityReport` resource, the user is not expected to add custom labels to it.

I would just set the labels we expect. We can always change that in the future if there's actual demand from users.

BTW, we were to keep the user labels around, a better way to do that would have been:

```golang
if vulnerabilityReport.Labels == nil {
  vulnerabilityReport.Labels = make(map[string]string)
}

vulnerabilityReport.Labels[constants.LabelManagedByKey] = constants.LabelManagedByValue
```

flavio:
@pohanhuangtw don't forget to rebase against main, the updates of golangci-lint caused some merge conflicts

helm/templates/_helpers.tpl
```diff
@@ -39,7 +39,7 @@ helm.sh/chart: {{ include "sbombastic.chart" . }}
 {{- if .Chart.AppVersion }}
 app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
 {{- end }}
-app.kubernetes.io/managed-by: {{ .Release.Service }}
```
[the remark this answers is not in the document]
pohanhuangtw: I think this has made into the main branch, so I will skip this

internal/handlers/scan_sbom.go
```diff
@@ -41,6 +41,46 @@ func NewScanSBOMHandler(
...
+		if err = file.Close(); err != nil {
+			h.logger.Error("failed to close temporary report file", "error", err)
+		}
+
+		if err = os.Remove(file.Name()); err != nil {
+			h.logger.Error("failed to remove temporary report file", "error", err)
+		}
+	}
+	return file, cleanup, nil
+}
```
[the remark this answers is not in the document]
pohanhuangtw: Sure, let me revert it

flavio:
looks good now, please squash all the commits into a single one, then we can merge it

pohanhuangtw:
Hi @flavio ,
Just squashed into 1 commit, please take a look, thanks

PR #181, opened 2025-04-23
feat: add e2e test for sbombastic

pohanhuangtw:
## Description

- Fix https://github.com/rancher-sandbox/sbombastic/issues/32
- Support 
  - e2e test in GHA, every 24 hour
  - able to run locally
  - create cluster, namespace, install by helm with local build controller / worker / storage

## Test

- run `make test-e2e`

### TODO
- [x] release GHA, because we lack of the release flow now
- [ ] image will not be delete if registry CR is deleted, track in this [issue](https://github.com/rancher-sandbox/sbombastic/issues/182).

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,0 +1,15 @@
+name: E2E test
+
+on:
+  pull_request:
+
+jobs:
+  test:
+    name: Run E2E tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - uses: actions/setup-go@0aaccfd150d50ccaeb58ebd88d36e91967a5f35b # v5.4.0
+        with:
+          go-version: "1.24.1"
+      - run: make test-e2e
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,13 @@ ENVTEST_K8S_VERSION := 1.31.0
 CONTROLLER_GEN ?= go run sigs.k8s.io/controller-tools/cmd/controller-gen@$(CONTROLLER_TOOLS_VERSION)
 ENVTEST ?= go run sigs.k8s.io/controller-runtime/tools/setup-envtest@$(ENVTEST_VERSION)
 
+GO_MOD_SRCS := go.mod go.sum
+
 ENVTEST_DIR ?= $(shell pwd)/.envtest
+
+REGISTRY ?= ghcr.io
+REPO ?= rancher-sandbox/sbombastic
+TAG ?= latest
 
 .PHONY: all
 all: controller storage worker
@@ -17,6 +23,10 @@ test: vet ## Run tests.
 .PHONY: helm-unittest
 helm-unittest:
 	helm unittest helm/ --file "tests/**/*_test.yaml"
+
+.PHONY: test-e2e
+test-e2e: controller-image storage-image worker-image
+	go test ./test/e2e/ -v
 
 .PHONY: fmt
 fmt:
@@ -34,17 +44,44 @@ lint-fix: golangci-lint ## Run golangci-lint linter and perform fixes
 vet:
 	go vet ./...
 
-.PHONY: controller
-controller: vet
+CONTROLLER_SRC_DIRS := cmd/controller api internal/controller
+CONTROLLER_GO_SRCS := $(shell find $(CONTROLLER_SRC_DIRS) -type f -name '*.go')
+CONTROLLER_SRCS := $(GO_MOD_SRCS) $(CONTROLLER_GO_SRCS)
+.PHONY: controller 
+controller: $(CONTROLLER_SRCS) vet
 	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o ./bin/controller ./cmd/controller
 
-.PHONY: storage
-storage: vet
-	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o ./bin/storage ./cmd/storage 
+.PHONY: controller-image
+controller-image:
+	docker build -f ./Dockerfile.controller \
+		-t "$(REGISTRY)/$(REPO)/controller:$(TAG)" .
+	@echo "Built $(REGISTRY)/$(REPO)/controller:$(TAG)"
 
+STORAGE_SRC_DIRS := cmd/storage api internal/apiserver internal/storage pkg
+STORAGE_GO_SRCS := $(shell find $(STORAGE_SRC_DIRS) -type f -name '*.go')
+STORAGE_SRCS := $(GO_MOD_SRCS) $(STORAGE_GO_SRCS)
+.PHONY: storage
+storage: $(STORAGE_SRCS) vet
+	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o ./bin/storage ./cmd/storage
+
+.PHONY: storage-image
+storage-image:
+	docker build -f ./Dockerfile.storage \
+		-t "$(REGISTRY)/$(REPO)/storage:$(TAG)" .
+	@echo "Built $(REGISTRY)/$(REPO)/storage:$(TAG)"
+
+WORKER_SRC_DIRS := cmd/worker api internal/messaging internal/handlers
+WORKER_GO_SRCS := $(shell find $(WORKER_SRC_DIRS) -type f -name '*.go')
+WORKER_SRCS := $(GO_MOD_SRCS) $(WORKER_GO_SRCS)
 .PHONY: worker
-worker: vet
+worker: $(WORKER_SRCS) vet
 	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o ./bin/worker ./cmd/worker
+
+.PHONY: worker-image
+worker-image:
+	docker build -f ./Dockerfile.worker \
+		-t "$(REGISTRY)/$(REPO)/worker:$(TAG)" .
+	@echo "Built $(REGISTRY)/$(REPO)/worker:$(TAG)"
 
 .PHONY: generate
 generate: generate-controller generate-storage generate-mocks
--- a/helm/templates/storage/deployment.yaml
+++ b/helm/templates/storage/deployment.yaml
@@ -22,15 +22,20 @@ spec:
       containers:
         - name: storage
           image: {{ .Values.storage.image.repository }}:{{ .Values.storage.image.tag }}
-          {{- if .Values.storage.logLevel }}
           args:
+            - --cert-dir=/certs
+          {{- if .Values.storage.logLevel }}
             - -log-level={{ .Values.storage.logLevel }}
           {{- end }}
           imagePullPolicy: {{ .Values.storage.image.pullPolicy }}
           volumeMounts:
             - name: sqlite-pvc
               mountPath: /data/sqlite/
+            - name: cert-volume
+              mountPath: /certs
       volumes:
         - name: sqlite-pvc
           persistentVolumeClaim:
             claimName: {{ if .Values.persistence.storageData.existingClaim }}{{ .Values.persistence.storageData.existingClaim }}{{- else }}{{ template "sbombastic.fullname" . }}-storage-data{{- end }}
+        - name: cert-volume
+          emptyDir: {}
--- a/test/e2e/e2e_suite_test.go
+++ b/test/e2e/e2e_suite_test.go
@@ -1,0 +1,121 @@
+package e2e
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/e2e-framework/klient/wait"
+	"sigs.k8s.io/e2e-framework/klient/wait/conditions"
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+	"sigs.k8s.io/e2e-framework/pkg/features"
+	"sigs.k8s.io/e2e-framework/third_party/helm"
+
+	storagev1alpha1 "github.com/rancher/sbombastic/api/storage/v1alpha1"
+	v1alpha1 "github.com/rancher/sbombastic/api/v1alpha1"
+)
+
+func EqualReference(img storagev1alpha1.ImageMetadata, registryURI, registryRepository, tag string) bool {
+	return img.RegistryURI == registryURI &&
+		img.Repository == registryRepository &&
+		img.Tag == tag
+}
+
+func TestRegistryCreation(t *testing.T) {
+	releaseName := "sbombastic"
+	registryName := "test-registry"
+	registryURI := "ghcr.io"
+	registryRepository := "rancher-sandbox/sbombastic/test-assets/golang"
+
+	crName := "dfe56d8371e7df15a3dde25c33a78b84b79766de2ab5a5897032019c878b5932"
+
+	f := features.New("Start Registry CR Creation test").
+		Setup(func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+			manager := helm.New(cfg.KubeconfigFile())
+			err := manager.RunInstall(helm.WithName(releaseName),
+				helm.WithNamespace(cfg.Namespace()),
+				helm.WithChart("../../helm"),
+				helm.WithWait(),
+				helm.WithArgs("--set", "controller.image.tag=latest",
+					"--set", "storage.image.tag=latest",
+					"--set", "worker.image.tag=latest"),
+				helm.WithTimeout("3m"))
+
+			require.NoError(t, err, "sbombastic helm chart is not installed correctly")
+
+			err = storagev1alpha1.AddToScheme(cfg.Client().Resources(cfg.Namespace()).GetScheme())
+			require.NoError(t, err)
+
+			err = v1alpha1.AddToScheme(cfg.Client().Resources(cfg.Namespace()).GetScheme())
+			require.NoError(t, err)
+
+			return ctx
+		}).
+		Assess("Create Registry CR", func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+			registry := &v1alpha1.Registry{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      registryName,
+					Namespace: cfg.Namespace(),
+				},
+				Spec: v1alpha1.RegistrySpec{
+					URI:          registryURI,
+					Repositories: []string{registryRepository},
+				},
+			}
+			err := cfg.Client().Resources().Create(ctx, registry)
+			require.NoError(t, err)
+			return ctx
+		}).
+		Assess("Verify the Image is created", func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+			images := storagev1alpha1.ImageList{
+				Items: []storagev1alpha1.Image{
+					{ObjectMeta: metav1.ObjectMeta{Name: crName, Namespace: cfg.Namespace()}},
+				},
+			}
+
+			err := wait.For(conditions.New(cfg.Client().Resources()).ResourcesFound(&images))
+			if err != nil {
+				t.Error(err)
+			}
+			return ctx
+		}).
+		Assess("Verify the SPDX SBOM is created", func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+			sboms := storagev1alpha1.SBOMList{
+				Items: []storagev1alpha1.SBOM{
+					{ObjectMeta: metav1.ObjectMeta{Name: crName, Namespace: cfg.Namespace()}},
+				},
+			}
+
+			err := wait.For(conditions.New(cfg.Client().Resources()).ResourcesFound(&sboms))
+			if err != nil {
+				t.Error(err)
+			}
+			return ctx
+		}).
+		Assess("Verify the VulnerabilityReport is created", func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+			vulnReports := storagev1alpha1.VulnerabilityReportList{
+				Items: []storagev1alpha1.VulnerabilityReport{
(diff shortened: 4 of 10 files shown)


holyspectral:
Looks neat!  A quick comment is that, have you considered using [wait.For](https://pkg.go.dev/sigs.k8s.io/e2e-framework/klient/wait#For) instead of assert.Eventually?  It would require the name of CRD to be deterministic, but it should be easy to achieve in test scenarios.  This would make your test code much simpler without having to care about k8s client.

For example, this is how we monitor if a CRD is created and process data is inserted: 

```
				proposal := v1alpha1.WorkloadSecurityPolicyProposal{
					ObjectMeta: v1.ObjectMeta{
						Name:      id,
						Namespace: "default",
					},
				}

				err := wait.For(conditions.New(r).ResourceMatch(&proposal, func(object k8s.Object) bool {
					obj := object.(*v1alpha1.WorkloadSecurityPolicyProposal)

					return slices.Contains(obj.Spec.Rules.Executables.Allowed, "/usr/bin/sleep")
				}), wait.WithTimeout(time.Minute*3))
				require.NoError(t, err)
```

Also you can bring custom data to next stage by using `context.WithValue(ctx, key("group"), id)`, and retrieve using `id := ctx.Value(key("group")).(string)`, so you don't have to re-initialize everything again.  This is also useful when you can't introduce global variables under the new linter rules.

holyspectral:
@pohanhuangtw yeah that's fine.  Just wanted to point out that as you write more tests, there are ways to save some time.  :-)

test/e2e/main_test.go
```diff
@@ -0,0 +1,58 @@
...
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+	"sigs.k8s.io/e2e-framework/pkg/envfuncs"
+	"sigs.k8s.io/e2e-framework/support/kind"
+)
+
+var (
+	testenv         env.Environment
+	kindClusterName string
+	namespace       string
+	workerImage     = "ghcr.io/rancher-sandbox/sbombastic/worker:v0.1.0-alpha1"
```
holyspectral, line 33: Do you have to use `v0.1.0-alpha1` tag here?  It's probably alright in most cases, but I'd personally prefer to have a tag that won't exist in any case (like `dev` or `latest` if you know `latest` will not be used).  This way there is no chance a wrong image will be used accidentally.

test/e2e/e2e_suite_test.go
```diff
@@ -0,0 +1,264 @@
...
+						return EqualReference(
+							sbomObj.Spec.ImageMetadata,
+							registryURI,
+							registryRepository,
+							golangAlpineTag,
+						)
+					}),
+				wait.WithInterval(pollInterval),
+				wait.WithTimeout(pollTimeout),
+			)
```
holyspectral, line 146: I understand that I probably have a similar code in my e2e tests lol, but generally if your SBOM resource is created with all information filled and it will not be changed later, you can just wait for the resource to be created and check its content. This way, in case of failure, you don't have to wait for the whole timeout to fail the test.  

```suggestion
			var sbomObj *storagev1alpha1.SBOM
			err := wait.For(
				conditions.New(cfg.Client().Resources(cfg.Namespace())).ResourceMatch(
					&sbom,
					func(obj k8s.Object) bool {
						var ok bool
						sbomObj, ok = obj.(*storagev1alpha1.SBOM)
						if !ok {
							t.Fatal("unexpected type assertion failure")
						}
						return true
					}),
				wait.WithInterval(pollInterval),
				wait.WithTimeout(pollTimeout),
			)
			require.NoError(t, err)
			assert.True(t, EqualReference(
				sbomObj.Spec.ImageMetadata,
				registryURI,
				registryRepository,
				golangAlpineTag,
			))
			require.NoError(t, err, "SBOM CR did not reach expected state within %s", pollTimeout)
```

holyspectral (changes requested):
I mark this as request for change for the hardcoded path in helm charts.  Other than that nothing is major.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,16 @@
+name: E2E test
+
+on:
+  schedule:
+    - cron: "0 0 * * *"
```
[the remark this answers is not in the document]
pohanhuangtw: Currently each run takes within 10 mins, is it a good practice to run in each PR is merged?

test/e2e/main_test.go
pohanhuangtw: You are right about this concern.
Even though the test will prioritize loading the local image, the tag should still be taken into consideration.

.github/workflows/e2e.yml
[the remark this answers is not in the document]
pohanhuangtw: @flavio what do you think about this?

Makefile
```diff
@@ -9,6 +9,27 @@ ENVTEST ?= go run sigs.k8s.io/controller-runtime/tools/setup-envtest@$(ENVTEST_V
...
+build-e2e-$(1)-image:
+	$(IMAGE_BUILDER) build -f ./Dockerfile.$(1) \
+	-t "$(REGISTRY)/$(REPO)/sbombastic/$(1):$(E2E_TAG)" $(BUILD_ACTION) .
+	@echo "Built $(REGISTRY)/$(REPO)/sbombastic/$(1):$(E2E_TAG)"
+
+E2E_DEPS += build-e2e-$(1)-image
+endef
+
+TARGETS=controller storage worker
+$(foreach target,$(TARGETS),$(eval $(call BUILD_template,$(target))))
```
holyspectral, line 31: These targets will be used by SLSA build as well.  Could you keep this part as your previous commit but override `$(TAG)` when running `make $(E2E_DEPS)`?

holyspectral (changes requested):
Generally looks good to me.  Just a tiny issue about how e2e images are built.

Makefile
pohanhuangtw: Sure, Thanks for the help.

Makefile
```diff
@@ -9,6 +9,27 @@ ENVTEST ?= go run sigs.k8s.io/controller-runtime/tools/setup-envtest@$(ENVTEST_V
...
+	$(IMAGE_BUILDER) build -f ./Dockerfile.$(1) \
+	-t "$(REGISTRY)/$(REPO)/sbombastic/$(1):$(TAG)" $(BUILD_ACTION) .
+	@echo "Built $(REGISTRY)/$(REPO)/sbombastic/$(1):$(TAG)"
+
+E2E_DEPS += build-e2e-$(1)-image
+endef
+
+TARGETS=controller storage worker
+$(foreach target,$(TARGETS),$(eval $(call BUILD_template,$(target))))
+
```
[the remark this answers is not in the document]
pohanhuangtw: Just apply, thanks. Will used suggestion next time, forget to use this.

holyspectral (approved):
LGTM

test/e2e/e2e_suite_test.go
```diff
@@ -0,0 +1,273 @@
...
+				{resource: &sbom, kind: "SBOM"},
+				{resource: &vulnReport, kind: "VulnerabilityReport"},
+			} {
+				require.NoError(t, wait.For(
+					conditions.New(cfg.Client().Resources(cfg.Namespace())).ResourceDeleted(
+						obj.resource),
+					wait.WithInterval(pollInterval),
+					wait.WithTimeout(pollTimeout),
+				), "%v CR was not deleted after Registry CR was deleted", obj.kind)
+			}
```
flavio, line 261: It's not a good idea to structure the code that waits for the removal in this way, that's because all these objects have parent - child relations.

```mermaid
classDiagram
        SBOM <|-- VulnerabilityReport
        Image <|-- SBOM
```

These relations end up determining the deletion order. Meaning that the resources will be deleted in this order:

1. VulnerabilityReport
2. SBOM
3. Image
4. Registry

The `for` loop starts by waiting for `Image` to be removed, but the `wait` might fail if the deletion of `VulnerabilityReport` and `SBOM` is not fast enough.

Get rid of the `for` loop and write individual checks like the following one:

```golang
err = wait.For(
  conditions.New(
   cfg.Client().ResourceDeleted(vulnReport),
   wait.WithTimeout(2*time.Minute),
   wait.WithImmediate(),
  )
if err != nil {
  t.Error("failed waiting for VulnerabilityReport resource to be deleted", err)
}
```

Some additional notes:
- Specifying the Namespace to be used is not required, since this is already part of the object given to the `ResourceDeleted`


Thinking more, I think we can even rewrite the code I proposed above to use label selectors, like shown [here](https://github.com/kubernetes-sigs/e2e-framework/tree/main/examples/wait_for_resources#waiting-for-a-lists-of-objects) (see the "check for their absence").
This is what I would do:

- get all the VulnerabilityReport/Image/SBOM associated with the Registry
- trigger the delete of the registry
- wait for the list of objects to be deleted

Makefile
```diff
@@ -71,3 +92,11 @@ generate-storage: generate-storage-test-crd ## Generate storage  code in pkg/gen
...
 generate-mocks: ## Generate mocks for testing.
 	go generate ./...
+
+.PHONY: test-e2e
+test-e2e:
+ifeq ($(E2E_NO_REBUILD),)
+	make TAG=e2e-test $(E2E_DEPS)
+endif
+	go test ./test/e2e/ -v
+
```
flavio, line 102: I think this is a bit more cryptic, plus we have to resort to some variable and human decision to decide whether or not the container images have to be rebuilt.

We only have 3 container images, all the core binaries of them are built by dedicated Makefile targets. If we make the source files of each binary known to the Makefile, then we can have conditional compilation done by Makefile for the binaries **and** for their container images.
The list of dependencies of these binaries is already specified inside of the `Tiltfile`.

I would:

- Define 3 new targets, one per container image that has to be built
- Each container image target depends on the core binary
- Each code binary target specifies which source files it depends on
- the `test-e2e` test depends on the container image targets

flavio (changes requested):

helm/templates/storage/deployment.yaml
```diff
@@ -22,14 +22,16 @@ spec:
       containers:
         - name: storage
           image: {{ .Values.storage.image.repository }}:{{ .Values.storage.image.tag }}
-          {{- if .Values.storage.logLevel }}
           args:
+            - "--cert-dir=/tmp/apiserver.local.config/certificates/"
```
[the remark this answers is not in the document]
pohanhuangtw: When running the storage API server, it requires a writable location to store the generated certificates.
By default, the pod does not have permission to write to the default certificate directory, which results in the following error:

```bash
E0418 06:13:50.437901       1 run.go:72] "command failed" err="error creating self-signed certificates: mkdir apiserver.local.config: permission denied"
```

To address this, I updated the implementation to use a writable fallback path.
This approach was inspired by [Rancher's chart implementation](https://github.com/rancher/partner-charts/blob/f5dc74e498e752e9574a6427d7a4755d66784816/charts/kasten/k10/7.0.1/templates/_k10_container.tpl#L27).

test/e2e/e2e_suite_test.go
```diff
@@ -0,0 +1,273 @@
...
+			require.NoError(t, err)
+
+			// Filter out "DocumentNamespace" and any field named "AnnotationDate" or "Created" regardless of nesting,
+			// since they contain timestamps and are not deterministic.
+			filter := cmp.FilterPath(func(path cmp.Path) bool {
+				lastField := path.Last().String()
+				return lastField == ".DocumentNamespace" || lastField == ".AnnotationDate" || lastField == ".Created"
+			}, cmp.Ignore())
+			diff := cmp.Diff(expectedSPDX, generatedSPDX, filter, cmpopts.IgnoreUnexported(spdx.Package{}))
+			assert.Empty(t, diff)
```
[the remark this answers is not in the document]
pohanhuangtw: You're right — I think it's more important that the overall process makes sense not the content.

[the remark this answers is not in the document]
fabriziosestito: Yes I agree with flavio. We could check the actual presence of the resources and if they have some metadata (such as the right labels) in place. This is enough cause we already cover the rest via the integration tests.

.github/workflows/e2e.yml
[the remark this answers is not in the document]
fabriziosestito: I agree with @holyspectral and @flavio

fabriziosestito (changes requested):

test/e2e/e2e_suite_test.go
pohanhuangtw: Hi @flavio @fabriziosestito,

I think using a label selector to retrieve all VulnerabilityReport, Image, and SBOM resources associated with a Registry makes much more sense.

However, based on the current implementation, we don’t include registry information in the resource labels.
Would it make sense for me to include that in the current PR?

For reference, [PR #176](https://github.com/rancher-sandbox/sbombastic/pull/176) shows the labels we're currently adding.

pohanhuangtw:
Hi @flavio @fabriziosestito @holyspectral,

Thanks for the review. Here's a summary of the changes included in the commits:

1. **Makefile**: Refactored to rebuild images only when relevant source files have changed.
2. **E2E tests**: Cleaned up by removing:
   - Redundant variables
   - Unnecessary content checks
   - Manual deletion logic (now using label-based deletion with labels added to the relevant CRs)
3. **Tiltfile**: Removed a non-existent dependency in the storage service
4. **Logging config in values**: 
   - Commented out the `TODO` for storage log level configuration to prevent YAML parsing errors.
   - added `certDir` to the API server for proper certificate writing.

flavio:
As discussed during the daily, let's remove the e2e tests that deals with the garbage collector and the owner reference.

This has to be added to our integration tests. This is already tracked via https://github.com/rancher-sandbox/sbombastic/issues/182

test/e2e/main_test.go
```diff
@@ -0,0 +1,58 @@
...
+	testenv.Setup(
+		envfuncs.CreateCluster(kind.NewProvider(), kindClusterName),
+		envfuncs.CreateNamespace(namespace),
+		envfuncs.LoadImageToCluster(kindClusterName, workerImage, "--verbose", "--mode", "direct"),
+		envfuncs.LoadImageToCluster(kindClusterName, controllerImage, "--verbose", "--mode", "direct"),
+		envfuncs.LoadImageToCluster(kindClusterName, storageImage, "--verbose", "--mode", "direct"),
+	)
+
+	testenv.Finish(
+		envfuncs.DeleteNamespace(namespace),
```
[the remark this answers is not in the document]
flavio: on the next line we call:

```golang
envfuncs.Destroycluster(kindClusterName)
```

The deletion of the namespace might require some time, depending on how much time it takes to terminate the pods running inside of it.
I was suggesting to not wait this time, given we immediately destroy the cluster (and all its contents) on the very next line

test/e2e/e2e_suite_test.go
```diff
@@ -0,0 +1,138 @@
...
+					Namespace: cfg.Namespace(),
+				},
+			}
+
+			err := wait.For(conditions.New(cfg.Client().Resources()).ResourceMatch(
+				&image,
+				func(_ k8s.Object) bool {
+					return true
+				}),
+				wait.WithImmediate())
```
flavio, line 86: This is not following what I recommended [here](https://github.com/rancher-sandbox/sbombastic/pull/181#discussion_r2077228695).

I was expecting to see something similar to [the 1st example](https://github.com/kubernetes-sigs/e2e-framework/tree/main/examples/wait_for_resources#waiting-for-a-single-object):

```golang
	var err error
	pod := v1.Pod{ObjectMeta: metav1.ObjectMeta{Name: "my-pod"}}
	err = wait.For(conditions.New(client.Resources()).PodRunning(pod), WithImmediate())
	if err != nil {
		t.Error(err)
	}
```

test/e2e/e2e_suite_test.go
```diff
@@ -0,0 +1,138 @@
...
+					Namespace: cfg.Namespace(),
+				},
+			}
+
+			err := wait.For(conditions.New(cfg.Client().Resources()).ResourceMatch(
+				&sbom,
+				func(_ k8s.Object) bool {
+					return true
+				}),
+				wait.WithImmediate())
```
flavio, line 104: Same here

test/e2e/e2e_suite_test.go
```diff
@@ -0,0 +1,138 @@
...
+					Namespace: cfg.Namespace(),
+				},
+			}
+
+			err := wait.For(conditions.New(cfg.Client().Resources()).ResourceMatch(
+				&vulnReport,
+				func(_ k8s.Object) bool {
+					return true
+				}),
+				wait.WithImmediate())
```
flavio, line 122: Same here

test/e2e/e2e_suite_test.go
flavio: > However, based on the current implementation, we don’t include registry information in the resource labels.
> Would it make sense for me to include that in the current PR?

No, it's not related with this PR and with what we're trying to achieve.

We can add it if there's going to be a use case for that. The deletion of all the resources associated with a Registry is not a good reason for introducing it, since we should just delete the Registry and let the Kubernete's GC handle all the other removals for us.

> For reference, https://github.com/rancher-sandbox/sbombastic/pull/176 shows the labels we're currently adding.

The PR is about adding [k8s's recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/). This label you're talking about, if we will add it, is not related with them.

Makefile
flavio: > - Define 3 new targets, one per container image that has to be built
> - Each container image target depends on the core binary

This is still missing

flavio (changes requested):

test/e2e/e2e_suite_test.go
pohanhuangtw: The source code behind [`[PodRunning](https://github.com/kubernetes-sigs/e2e-framework/blob/ff39024351aaf71584de1f2e2ca0b94f803d8570/klient/wait/conditions/conditions.go#L281)`](https://github.com/kubernetes-sigs/e2e-framework/blob/ff39024351aaf71584de1f2e2ca0b94f803d8570/klient/wait/conditions/conditions.go#L281) calls:

```go
c.resources.Get(context.Background(), obj.GetName(), obj.GetNamespace(), obj)
```

This is similar to what the current implementation is doing.

Another relevant API is:

```go
func (c *Condition) ResourcesFound(list k8s.ObjectList) apimachinerywait.ConditionWithContextFunc {
	return c.ResourcesMatch(list, func(object k8s.Object) bool { return true })
}
```

It also eventually calls [`[ResourceMatch](https://github.com/kubernetes-sigs/e2e-framework/blob/ff39024351aaf71584de1f2e2ca0b94f803d8570/klient/wait/conditions/conditions.go#L112)`](https://github.com/kubernetes-sigs/e2e-framework/blob/ff39024351aaf71584de1f2e2ca0b94f803d8570/klient/wait/conditions/conditions.go#L112).

So, I believe the current implementation is reasonable and consistent with what the framework does internally.

test/e2e/e2e_suite_test.go
pohanhuangtw: same here

test/e2e/e2e_suite_test.go
pohanhuangtw: same here

internal/handlers/constants.go
```diff
@@ -5,4 +5,5 @@ const (
 	LabelManagedByValue = "sbombastic"
 	LabelPartOfKey      = "app.kubernetes.io/part-of"
 	LabelPartOfValue    = "sbombastic"
+	LabelRegistryKey    = "registry"
```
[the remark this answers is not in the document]
pohanhuangtw: You are right, I try to get the image/SBOM/report CR by reading the labels, but I found this is not redundant, I will remove it

pohanhuangtw:
The conversation is getting long, so here's a quick summary of the latest changes across the last few commits:

1. **Removed** the `registry` label.
2. **Removed** the call to `envfuncs.DeleteNamespace(namespace)`.
3. **Removed** the CLI configuration for the cert directory—it's now hardcoded and backed by an `emptyDir`.

---

### In response to:

> * Define 3 new targets, one per container image that has to be built
> * Each container image target depends on the core binary

This has been addressed [here](https://github.com/rancher-sandbox/sbombastic/pull/181/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R58-R81).

With this setup, we ensure that each image gets rebuilt only when its corresponding core binary changes. Example:

```makefile
.PHONY: test-e2e
test-e2e: controller storage worker
	make TAG=e2e-test $(E2E_DEPS)
	go test ./test/e2e/ -v
```

---

### Regarding: [discussion link](https://github.com/rancher-sandbox/sbombastic/pull/181#discussion_r2084354598)

The source code behind [PodRunning](https://github.com/kubernetes-sigs/e2e-framework/blob/ff39024351aaf71584de1f2e2ca0b94f803d8570/klient/wait/conditions/conditions.go#L281) calls:

```go
c.resources.Get(context.Background(), obj.GetName(), obj.GetNamespace(), obj)
```

This is consistent with what the current implementation already does.

Another relevant function is:

```go
func (c *Condition) ResourcesFound(list k8s.ObjectList) apimachinerywait.ConditionWithContextFunc {
	return c.ResourcesMatch(list, func(object k8s.Object) bool { return true })
}
```

Which also eventually delegates to [ResourceMatch](https://github.com/kubernetes-sigs/e2e-framework/blob/ff39024351aaf71584de1f2e2ca0b94f803d8570/klient/wait/conditions/conditions.go#L112).

Based on that, I believe our current logic aligns well with how the e2e framework expects these checks to work.

---

Let me know if you see any gaps or areas for improvement.

flavio (changes requested):
I've done some refactor to the `Makefile`, see 227c21b558afe1954f094d73b7a1309744f26c6e.

The docker caching system works pretty well, the container images will be rebuilt only when there's an actual change to the relevant sources. That means that running `make test-e2e` multiple times is not going to lead to the rebuild of the images unless this is really needed. I've tried it locally

pohanhuangtw:
Hi @flavio ,
Just update the go mod and requirement from your end, please help take a look, thanks

flavio (commented):
LGTM now, thanks for having done all the requested changes. Can you now squash everything into a single commit?

flavio (approved):

PR #207, opened 2025-05-15
fix: add restOptions to ensure the garbage collection is enabled

pohanhuangtw:
## Description

- Fix https://github.com/rancher-sandbox/sbombastic/issues/182
- Introduce RestOptionsGetter to enable EnableGarbageCollection. 

Fix #XXX

## Test

### Potential improvement
Currently, Sbombastic is using SQLite. I'm not sure if we should address the parallel write [issue](https://github.com/mattn/go-sqlite3/issues/1022#issuecomment-1067353980) at this point. 

The common workaround is to limit the number of open connections by calling db.SetMaxOpenConns(1). However, I'm unsure whether this change belongs in this PR.

### DIFF
--- a/internal/apiserver/apiserver.go
+++ b/internal/apiserver/apiserver.go
@@ -110,17 +110,18 @@ func (c completedConfig) New(db *sqlx.DB, logger *slog.Logger) (*WardleServer, e
 
 	apiGroupInfo := genericapiserver.NewDefaultAPIGroupInfo(v1alpha1.GroupName, Scheme, metav1.ParameterCodec, Codecs)
 
-	imageStore, err := storage.NewImageStore(Scheme, c.GenericConfig.RESTOptionsGetter, db, logger)
+	restOptionsGetter := GetRESTOptionsGetter()
+	imageStore, err := storage.NewImageStore(Scheme, restOptionsGetter, db, logger)
 	if err != nil {
 		return nil, fmt.Errorf("error creating Image store: %w", err)
 	}
-	sbomStore, err := storage.NewSBOMStore(Scheme, c.GenericConfig.RESTOptionsGetter, db, logger)
+	sbomStore, err := storage.NewSBOMStore(Scheme, restOptionsGetter, db, logger)
 	if err != nil {
 		return nil, fmt.Errorf("error creating SBOM store: %w", err)
 	}
 	vulnerabilityReportStore, err := storage.NewVulnerabilityReport(
 		Scheme,
-		c.GenericConfig.RESTOptionsGetter,
+		restOptionsGetter,
 		db,
 		logger,
 	)
--- a/internal/apiserver/options.go
+++ b/internal/apiserver/options.go
@@ -1,0 +1,29 @@
+package apiserver
+
+import (
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apiserver/pkg/registry/generic"
+)
+
+type RestOptionsGetter struct {
+}
+
+// GetRESTOptions implements the generic.RESTOptionsGetter interface to ensure the GarbageCollection is enabled.
+// ResourcePrefix is set to "/<group>/<resource>", compatible with the storage Key format: /storage.sbombastic.rancher.io/<resource>/<namespace>/<name>.
+func (o *RestOptionsGetter) GetRESTOptions(
+	resource schema.GroupResource,
+	_ runtime.Object,
+) (generic.RESTOptions, error) {
+	return generic.RESTOptions{
+		EnableGarbageCollection: true,
+		DeleteCollectionWorkers: 1,
+		ResourcePrefix:          fmt.Sprintf("/%s/%s", resource.Group, resource.Resource),
+	}, nil
+}
+
+func GetRESTOptionsGetter() *RestOptionsGetter {
+	return &RestOptionsGetter{}
+}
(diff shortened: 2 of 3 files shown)


internal/apiserver/options.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apiserver/pkg/registry/generic"
+)
+
+type RestOptionsGetter struct {
```
fabriziosestito, line 11: this type can be be private

internal/apiserver/options.go
```diff
@@ -0,0 +1,29 @@
...
+	_ runtime.Object,
+) (generic.RESTOptions, error) {
+	return generic.RESTOptions{
+		EnableGarbageCollection: true,
+		DeleteCollectionWorkers: 1,
+		ResourcePrefix:          fmt.Sprintf("/%s/%s", resource.Group, resource.Resource),
+	}, nil
+}
+
+func GetRESTOptionsGetter() *RestOptionsGetter {
```
fabriziosestito, line 27: This isn't strictly necessary, since we can instantiate the type directly with `foo := &restOptionsGetter{}`.

test/e2e/e2e_suite_test.go
```diff
@@ -107,6 +105,70 @@ func TestRegistryCreation(t *testing.T) {
 			}
 			return ctx
 		}).
+		Assess("Verify the OwnerReference deletion", func(ctx context.Context, t *testing.T, cfg *envconf.Config) context.Context {
+			// Get all the VulnerabilityReport/Image/SBOM associated with the Registry
+			labelSelector := labels.FormatLabels(
+				map[string]string{handlers.LabelManagedByKey: handlers.LabelManagedByValue},
+			)
+			images := storagev1alpha1.ImageList{}
+			err := wait.For(conditions.New(cfg.Client().Resources()).ResourceListN(
```
fabriziosestito, line 114: We are using Testify for the other tests. We should use it here as well instead of relying on the idiom:
```
if err != nil {
  t.Error(err)
}
```

fabriziosestito: e.g.`require.NoEror(t, err)`

fabriziosestito (changes requested):
Good job, minor styling issues spotted.

flavio (commented):
Good catch, LGTM once the notes left by Fabrizio are addressed :clap:

internal/apiserver/apiserver.go
```diff
@@ -110,17 +110,18 @@ func (c completedConfig) New(db *sqlx.DB, logger *slog.Logger) (*WardleServer, e
 
 	apiGroupInfo := genericapiserver.NewDefaultAPIGroupInfo(v1alpha1.GroupName, Scheme, metav1.ParameterCodec, Codecs)
 
-	imageStore, err := storage.NewImageStore(Scheme, c.GenericConfig.RESTOptionsGetter, db, logger)
+	restOptionsGetter := GetRESTOptionsGetter()
```
fabriziosestito, line 113: On second thought, couldn't we just assign this in `c.GenericConfig.RESTOptionsGetter` at [[this line](https://github.com/rancher-sandbox/sbombastic/blob/df70856d3a26603c63db7498ec69f99b3a09dad9/cmd/storage/server/start.go#L217)](https://github.com/rancher-sandbox/sbombastic/blob/df70856d3a26603c63db7498ec69f99b3a09dad9/cmd/storage/server/start.go#L217) and leave the rest of the code unchanged?

fabriziosestito (changes requested):

internal/apiserver/apiserver.go
pohanhuangtw: Thanks, this is a cleaner thought, and it works! 
Will push changes here when https://github.com/rancher-sandbox/sbombastic/pull/208 is merge.

pohanhuangtw:
Hi @flavio @fabriziosestito ,
the update has been made, here are the summaries.
1. put the init rest in server config
2. improve the error handling / test case coverage in the e2e test.

test/e2e/e2e_suite_test.go
```diff
@@ -101,6 +101,51 @@ func TestRegistryCreation(t *testing.T) {
...
+			err = wait.For(conditions.New(cfg.Client().Resources()).ResourceDeleted(registry))
+			require.NoError(t, err)
+
+			// Wait for the VulnerabilityReport/Image/SBOM to be deleted
+			err = wait.For(conditions.New(cfg.Client().Resources()).ResourcesDeleted(&images))
+			require.NoError(t, err)
+			err = wait.For(conditions.New(cfg.Client().Resources()).ResourcesDeleted(&sboms))
+			require.NoError(t, err)
+			err = wait.For(conditions.New(cfg.Client().Resources()).ResourcesDeleted(&vulnReports))
+			require.NoError(t, err)
```
flavio, line 145: I would change the order to match how resources are garbage collected, just to avoid some timing issue.

We should use this order:
- Vulnerability report 
- SBOM
- Image
- Registry

flavio (commented):
LGTM, just a minor change left

test/e2e/e2e_suite_test.go
pohanhuangtw: Sure, let me work on it now.

pohanhuangtw: Will do!

flavio (approved):

fabriziosestito (approved):
LGTM!

PR #213, opened 2025-05-20
feat: harden deployment and improve e2e test

pohanhuangtw:
## Description

- Fix https://github.com/rancher-sandbox/sbombastic/issues/205
- Add the harden deployment in security context for worker / storage / contorller
- caps / privileges are drop
- all the components can run in the readonly env (which default run in such condition.)
- improve the e2e test to run it in harden environment.

## Test

- `make test-e2e`


helm/templates/controller/statefulset.yaml
```diff
@@ -55,6 +52,12 @@ spec:
             requests:
               cpu: 10m
               memory: 64Mi
+          volumeMounts:
+            - name: nats-dats
+              mountPath: /tmp
```
flavio, line 57: I think the name of the volume has a typo, plus so not set that to be `/tmp`, make it something specific to NATS please

helm/templates/worker/deployment.yaml
```diff
@@ -32,6 +34,10 @@ spec:
           volumeMounts:
             - mountPath: /var/run/worker
               name: run-volume
+            - mountPath: /tmp
+              name: worker-tmp
```
flavio, line 38: Are you sure we need to do that? I  don't have a way to verify that, but I thought this directory is always mounted as a tmpfs, regardless of the read only setting.

test/e2e/main_test.go
```diff
@@ -27,7 +27,14 @@ func TestMain(m *testing.M) {
...
 	testenv.Setup(
 		envfuncs.CreateCluster(kind.NewProvider(), kindClusterName),
-		envfuncs.CreateNamespace(namespace),
+		envfuncs.CreateNamespace(namespace, envfuncs.WithLabels(map[string]string{
+			"pod-security.kubernetes.io/enforce":         "restricted",
+			"pod-security.kubernetes.io/enforce-version": "latest",
+			"pod-security.kubernetes.io/audit":           "restricted",
+			"pod-security.kubernetes.io/audit-version":   "latest",
+			"pod-security.kubernetes.io/warn":            "restricted",
+			"pod-security.kubernetes.io/warn-version":    "latest",
```
flavio, line 36: We can drop these lines since we're enforcing the profile. If something is not compliant we will be immediately blocked, hence there's no need for the warning directive. Finally, when something is not compliant we will get an error on the stderr, making the audit irrelevant

flavio (changes requested):

helm/templates/worker/deployment.yaml
pohanhuangtw: Thanks for the review! 
Yes, we needs this to ensure the trivy is able to write the tmp.
Are you ok with this naming?

flavio: Maybe make it something like tmp-dir

helm/templates/worker/deployment.yaml
```diff
@@ -32,6 +34,10 @@ spec:
           volumeMounts:
             - mountPath: /var/run/worker
               name: run-volume
+            - mountPath: /tmp
+              name: tmp-dir
```
flavio, line 38: I just realized we've already a volume to be used by the worker data. Can we configure trivy to make use of this directory instead?

internal/handlers/generate_sbom.go
```diff
@@ -102,6 +103,11 @@ func (h *GenerateSBOMHandler) Handle(message messaging.Message) error {
 		),
 	})
 
+	err = os.Setenv("TMPDIR", h.workDir)
+	if err != nil {
+		return fmt.Errorf("failed to set TMPDIR: %w", err)
+	}
+
```
flavio, line 110: Are you doing that to force trivy to use this directory as a cache dir? If that's the case, let's avoid doing that.

trivy has a `--cache-dir` flag, let's use that instead.

flavio (changes requested):

internal/handlers/generate_sbom.go
pohanhuangtw: When Trivy performs a scan, it calls [os.MkdirTemp](https://github.com/aquasecurity/trivy/blob/fe127715e505d753e0d878d52c5f280cdc326b76/pkg/fanal/artifact/image/image.go#L67) to create a temporary directory. However, this process does not bring in any cached folder in it, which leads to the issue we’re encountering.

flavio: I see, then let's go back to the initial approach of mounting a volume also for `/tmp`

pohanhuangtw: I wil revert the commit then.

pohanhuangtw:
If the PR has no issue, can I squash all of the commits?

pohanhuangtw:
Hi @flavio 

If the PR has no issue, may I squash all of the commits?

flavio:
> Hi @flavio
> 
> If the PR has no issue, may I squash all of the commits?

Sure, do the squash, then I'll approve and merge it

fabriziosestito:
nit: @pohanhuangtw let's stat adding the right tags to the PR, including the `helm`/`storage`/`worker`/`controller` component tag depending what we touched

PR #224, opened 2025-05-29
refactor: remove the embedded NATS server

fabriziosestito:
## Description

This PR removes the embedded NATS server and uses the official Helm chart for NATS deployment. This change reduces our codebase and enables us to take advantage of features provided by the chart, such as automatic TLS certificate reloading.

The controller is now deployed as a `Deployment` instead of a `StatefulSet`, as persistent storage is no longer required, simplifying the overall installation. The internal NATS service has been removed: the controller and worker components now connect to the NATS server deployed via the Helm chart.

The embedded NATS server remains in use for integration tests.

## TODO

* Add TLS certificates using cert-manager
* Allow disabling the default NATS installation and let users specify a custom NATS server URL


fabriziosestito:
@flavio @pohanhuangtw This PR should also address the flakiness in the e2e tests. Previously, the tests were running with 3 controller replicas, but the NATS cluster wasn't properly formed. As a result, the service was routing requests randomly to 3 separate, unconnected NATS servers.

flavio (approved):
I'm super happy about these changes, they make me feel even more convinced about this design decision

flavio:
> Allow disabling the default NATS installation and let users specify a custom NATS server URL

IMHO this has a super low priority

PR #225, opened 2025-05-30
test(e2e): export cluster logs

fabriziosestito:
## Description

Currently, there is no mechanism to debug e2e test failures.
This PR adds support for exporting cluster logs during e2e test runs. 
When the CI workflow fails, the logs are also uploaded as a GitHub Actions artifact.


flavio (approved):
Nice improvement!

Issue #162, opened 2025-04-03
RFC: RegistryDiscovery


williamlin-suse:
Design and review/discussion about RegistryDiscovery feature

Issue #161, opened 2025-04-03
RegistryDiscovery for manual and scheduled discovery


williamlin-suse:
### Is your feature request related to a problem?

Currently when a Registry CR is reconciled by controller, the annotation `sbombastic.rancher.io/last-discovered-at` in 
the Registry CR is checked. Discovery operation is triggered when this annotation is empty. 
When controller reconciles a SBOM CR and sees the number of Image CRs is equal to the number of SBOM CRs for the Registry, annotation `sbombastic.rancher.io/last-discovered-at` is set as current time in the owner Registry CR.
It means no new image could be discovered anymore after a registry is discovered unless the annotation 
`sbombastic.rancher.io/last-discovered-at` is cleared in the Registry CR.

### Solution you'd like

Add a CRD `RegistryDiscovery` that a registry discovery operation could be triggered on demand or by schedule.
We also need to prevent new discovery operations from being triggered when there is an discovery operation running(per-registry).

PR #227, opened 2025-05-30
docs(rfc): add scan trigger RFC

fabriziosestito:
## Description

This PR introduces the scan trigger RFC, superseding the Scan flow of [RFC 0001](https://github.com/rancher-sandbox/sbombastic/blob/main/docs/rfc/0001_scanner_architecture_and_design.md) and eliminating the "discovery job" concept.

Supersedes #161 and #162.


flavio (approved):

flavio (dismissed):
I like the new design we came up with. The RFC is providing a good overview of how it works.

docs/rfc/0002_scan_trigger.md
```diff
@@ -0,0 +1,160 @@
...
+The `ScanJob` resource will include a status field to reflect the scan's progress and outcome. This status will contain:
+
+- `phase`: Indicates the current state of the scan (`Pending`, `InProgress`, `Complete`, `Failed`).
+- `conditions`: Represents detailed job conditions, similar to those used in Kubernetes Jobs, showing whether the scan completed successfully or encountered issues (`Complete`, `Failed`).
+- `imagesCount`: The number of images found in the registry during the scan.
+
+Example status:
+
+```yaml
+status:
```
pohanhuangtw, line 65: Could you help clarify the design behind the status field?
I understand that the `phase` field is the primary indicator of the current scan state and you've mentioned the expected transition through the scan flow.

However, the condition section is a bit confusing:

- Why are the separate status values under each condition when the phase already represent the overall state?
- What's the purpose of having the both Complete and Failed types present at the same time, both of them mark with same reason / message? Is it an expected condition? if so, what does it mean?
- Wouldn't this potentially create conflicting or redundant information? e.g. reason and message might not represent the similar message.

Could you elaborate on how condition are meant to be interpreted in relation to the phase?
Thanks!

docs/rfc/0002_scan_trigger.md
```diff
@@ -0,0 +1,160 @@
...
+This reduces the number of Kubernetes API interactions and streamlines the scanning workflow.
+
+## Error handling
+
+- Transient errors encountered during the scan process (such as network problems or registry downtime) will be automatically retried by both reconcilers and workers. Workers will use exponential backoff for these retries. If the scan continues to fail after multiple attempts, the `ScanJob` will be marked as `Failed` with a relevant error message.
+- For non-transient errors (like an invalid registry configuration), the `ScanJob` will be marked as `Failed` immediately, accompanied by a clear error message.
+
+## Registry deletion
+
+A finalizer will be added to the `Registry` resource to guarantee that deletion only proceeds once any ongoing `ScanJob` has either completed or failed.
```
pohanhuangtw, line 116: This is a brilliant design!

Even the scan flow has change, does the original owner reference still exist?
Registry -> image -> SBOM -> Vulnerability Report.

docs/rfc/0002_scan_trigger.md
```diff
@@ -0,0 +1,160 @@
...
+      message: "Job in progress"
+    - type: Failed
+      status: "False"
+      reason: "Processing
+      message: "Job in progress"
+```
+
+### Validation
+
+Only one `ScanJob` can run against a `Registry` at a time. If a `ScanJob` is already in progress for a `Registry`, creating another one will be rejected.
```
pohanhuangtw, line 81: When is the ScanJob CRD created? Is it triggered by the following situations?

1. Created registry CR?
2. manual scan?

Would be helpful to understand what triggers the creation of ScanJob, especially in the context of scan flow scope.

docs/rfc/0002_scan_trigger.md
```diff
@@ -0,0 +1,160 @@
...
+9. **The worker** sends a NATS message for each image to trigger SBOM generation.
+10. **Workers** generate SBOMs and send messages to initiate vulnerability scans.
+11. **Workers** create a `VulnerabilityReport` resource for each image with the scan results.
+12. **The `VulnerabilityReport` reconciler** monitors the number of `VulnerabilityReport` resources and, once it matches `ImagesCount`, marks the `ScanJob` as `Complete`.
+
+This design simplifies the architecture by retaining only the `ScanJob` and `VulnerabilityReport` reconcilers.
+Unlike the previous model, where the `Image` and `SBOM` reconcilers coordinated different stages of the scan, the worker now directly publishes follow-up jobs (e.g., SBOM generation, vulnerability scan) to the queue.
+This reduces the number of Kubernetes API interactions and streamlines the scanning workflow.
+
+## Error handling
```
pohanhuangtw, line 109: In Error Handling section, you mentioned that Worker can exponential backoff for these retries.
Just to clarify, does this imply the following:
1. worker is able to push the message to the NATs for later processing?
2. In cases like Trivy failing to download db, would that be treated as transient error and retried accordingly?

docs/rfc/0002_scan_trigger.md
```diff
@@ -0,0 +1,160 @@
...
+- For non-transient errors (like an invalid registry configuration), the `ScanJob` will be marked as `Failed` immediately, accompanied by a clear error message.
+
+## Registry deletion
+
+A finalizer will be added to the `Registry` resource to guarantee that deletion only proceeds once any ongoing `ScanJob` has either completed or failed.
+This ensures that scans are not interrupted mid-process, preserving the integrity of scan results and preventing orphaned resources.
+
+## Garbage collection
+
+A maximum of X ScanJob resources per registry will be retained in the system for auditing and historical purposes, with X being a configurable value.
```
pohanhuangtw, line 121: Could you elaborate the design behind this?
Does it imply that ScanJob CRs exceeding the limit X (per registry) will automatically delete the oldest one, once the new one is created?

pohanhuangtw (commented):
The overall design looks good to me. I've left a few comments for clarification and minor suggestions.

docs/rfc/0002_scan_trigger.md
flavio: > Even the scan flow has change, does the original owner reference still exist?

Yes, it still exists

docs/rfc/0002_scan_trigger.md
flavio: This is explained later in the document, under the ["Scheduled scans"](https://github.com/rancher-sandbox/sbombastic/blob/53dd89d5930d58fbdeeefdb26a679ef7c1f6b286/docs/rfc/0002_scan_trigger.md#scheduled-scans) section.

To summarize:
- Could be created manually by the user
- Are going to be created automatically by the internal cron handler

When a registry is created it will have its own refresh schedule (we can add one using a mutating admission controller if the user didn't provide one). The cron handler will schedule its refresh on his very next tick.

docs/rfc/0002_scan_trigger.md
flavio: > Does it imply that ScanJob CRs exceeding the limit X (per registry) will automatically delete the oldest one, once the new one is created?

Yes, that would be the case.

docs/rfc/0002_scan_trigger.md
pohanhuangtw: Thanks for the clarification. Now that the trigger paths are clear, what do you think about adding a concise **“ScanJob trigger matrix”** to the RFC—maybe right after the *Scheduled scans* section? Something like:

```
ScanJob can be triggered via:
• Manual request from a user (kubectl / API)
• Automatic cron refresh based on each Registry’s schedule
• One-shot refresh immediately after a Registry CR is created
```

Having the list in the document would let readers see, at a glance, every way a ScanJob can start, without having to piece it together from multiple paragraphs.

docs/rfc/0002_scan_trigger.md
fabriziosestito: 1. not pushing but `nacking` (negative acknowledgement) 
2. yes

pohanhuangtw: if nacking, does the job will re-queue back to nats?

fabriziosestito: yes, please refer to the NATS consumers documentation: https://docs.nats.io/nats-concepts/jetstream/consumers

PR #240, opened 2025-06-06
ci: add rancher annotation

pohanhuangtw:
## Description

- Fix https://github.com/rancher-sandbox/sbombastic/issues/235
- add annotation in chart
- add app-readme

### DIFF
--- a/helm/Chart.yaml
+++ b/helm/Chart.yaml
@@ -27,3 +27,22 @@ dependencies:
   - name: nats
     version: "1.3.7"
     repository: "https://nats-io.github.io/k8s/helm/charts/"
+
+annotations:
+  catalog.cattle.io/certified: rancher # Any application we are adding as a helm chart
+  catalog.cattle.io/namespace: cattle-sbombastic-system # Must prefix with cattle- and suffix with -system
+  catalog.cattle.io/release-name: rancher-sbombastic # If this is an upstream app, prefixing with rancher is the preferred naming choice.
+  catalog.cattle.io/ui-component: sbombastic # This is added for custom UI deployment of a chart
+  catalog.cattle.io/display-name: SBOMbastic # Only for Charts with custom UI
+  catalog.cattle.io/os: linux # this means linux only, other choice here is "windows". For charts that support both, don't add this annotation
+  # optional ones:
+  catalog.cattle.io/provides-gvr: "sbombastic.rancher.io/v1" # Declare that this chart provides a type, which other charts may use in `requires-gvr`. Only add to parent, not CRD chart.
+  # The following two will create a UI warning if the request is not available in cluster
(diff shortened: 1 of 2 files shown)


helm/app-readme.md
```diff
@@ -0,0 +1,23 @@
+# SBOMbastic
```
fabriziosestito, line 1: Is this really needed? it means we will need to keep the main README in sync

helm/Chart.yaml
```diff
@@ -27,3 +27,12 @@ dependencies:
   - name: nats
     version: "1.3.7"
     repository: "https://nats-io.github.io/k8s/helm/charts/"
+
+annotations:
```
flavio, line 31: Please, revisit the annotations you added with the ones used by kubewarden. See [here](https://github.com/kubewarden/helm-charts/blob/5cae27ee09eac6f9012685f4f4b530cb19b13b78/charts/kubewarden-controller/Chart.yaml#L30-L48)

helm/Chart.yaml
```diff
@@ -27,3 +27,12 @@ dependencies:
...
+  catalog.cattle.io/auto-install: SBOMbastic=0.1.0-alpha1
+  catalog.cattle.io/display-name: SBOMbastic
+  catalog.cattle.io/namespace: sbombastic
+  catalog.cattle.io/release-name: rancher-sbombastic
+  catalog.cattle.io/requests-cpu: "500m"
+  catalog.cattle.io/requests-memory: "128Mi"
```
flavio, line 37: We don't have any value inside of the helm chart, how did you come up with these numbers?

@fabriziosestito do you think they make sense?

helm/Chart.yaml
```diff
@@ -27,3 +27,22 @@ dependencies:
...
+
+annotations:
+  catalog.cattle.io/certified: rancher # Any application we are adding as a helm chart
+  catalog.cattle.io/namespace: cattle-sbombastic-system # Must prefix with cattle- and suffix with -system
+  catalog.cattle.io/release-name: rancher-sbombastic # If this is an upstream app, prefixing with rancher is the preferred naming choice.
+  catalog.cattle.io/ui-component: sbombastic # This is added for custom UI deployment of a chart
+  catalog.cattle.io/display-name: SBOMbastic # Only for Charts with custom UI
+  catalog.cattle.io/os: linux # this means linux only, other choice here is "windows". For charts that support both, don't add this annotation
+  # optional ones:
+  catalog.cattle.io/provides-gvr: "policyservers.policies.sbombastic.io/v1" # Declare that this chart provides a type, which other charts may use in `requires-gvr`. Only add to parent, not CRD chart.
```
flavio, line 39: this is wrong

flavio (changes requested):

fabriziosestito (approved):

PR #246, opened 2025-06-13
fix: install crane througth curl

pohanhuangtw:
## Description
- Fix https://github.com/rancher-sandbox/sbombastic/issues/245
- There is an issue when installing through kubewarden, so change it to curl install.
- Install with v0.20.5, verify with shasum

## Inspired 
- [RKE2](https://github.com/rancher/rke2/blob/c6c325dd17cbe12eb0a00d31bc8fdb8deeef5f80/.github/actions/install-crane/action.yml#L4)
- [kubewarden](https://github.com/kubewarden/github-actions/blob/6fa3010e6919516b8ba438eec3e6b8c5569a0f76/crane-installer/action.yml#L4
)

## Test

on fork mock [release](https://github.com/pohanhuangtw/sbombastic/actions/runs/15624733172).

### DIFF
--- a/.github/workflows/attestation.yml
+++ b/.github/workflows/attestation.yml
@@ -17,11 +17,23 @@ jobs:
       packages: write
       id-token: write
     runs-on: ubuntu-latest
+    env:
+      CRANE_VERSION: v0.20.5
+      CRANE_CHECKSUM: ad4cd9af2568c62c97e346de6d1295ee8c6ce3341f7b71cf02d41292b4532680
+
     steps:
       - name: Install cosign
         uses: sigstore/cosign-installer@3454372f43399081ed03b604cb2d021dabca52bb # v3.8.2
       - name: Install the crane command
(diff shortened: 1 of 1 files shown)


.github/workflows/attestation.yml
```diff
@@ -10,6 +10,16 @@ on:
         type: string
         description: "Architecture"
         required: true
+      version:
+        type: string
+        description: 'crane release to be installed'
+        required: false
+        default: v0.20.5
```
fabriziosestito, line 17: Maybe we could use env at job level in the "Install the crane command" step instead of modifying the action inputs.

e.g.

```
jobs:
  attest:
    runs-on: ubuntu-latest
    env:
      CRANE_VERSION: v0.20.5
      CRANE_CHECKSUM: ad4cd9af2568c62c97e346de6d1295ee8c6ce3341f7b71cf02d41292b4532680
      [...]
```

fabriziosestito (approved):

PR #270, opened 2025-06-20
feat: move helm to charts/sbombastic

pohanhuangtw:
## Description
- Fix https://github.com/rancher-sandbox/sbombastic/issues/255

## Test
- Verified that unit tests, integration tests, and end-to-end tests pass successfully.
- Deployed the stack using Tilt.
  - Applied a fake Registry Custom Resource (CR) and confirmed that both SBOM and VulnerabilityReport CRs were generated successfully.


pohanhuangtw:
When this PR is merge, will update the setting in https://github.com/rancher-sandbox/sbombastic/pull/251

fabriziosestito (changes requested):
The Makefile still needs work:

https://github.com/rancher-sandbox/sbombastic/blob/7ca18a3b2ea9cab778534a85244055074f37d81b/Makefile#L96

the manifest target is still generating resources in the wrong location

flavio (approved):

fabriziosestito (approved):

Issue #269, opened 2025-06-19
Add ScanJob reconciler


fabriziosestito:
Implement a `ScanJob` reconciler that follows the flow defined in [RFC 0002 - Triggering Scans](https://github.com/rancher-sandbox/sbombastic/blob/main/docs/rfc/0002_scan_trigger.md#triggering-scans-flow).

To avoid blocking the reconciler while publishing to the NATS queue, configure it with increased concurrency using the [`MaxConcurrentReconciles`](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller#TypedOptions) option. This is safe to use, as the underlying [workqueue implementation](https://github.com/kubernetes/client-go/tree/master/util/workqueue) guarantees that the same resource won't be processed concurrently by multiple goroutines.

PR #275, opened 2025-06-23
feat(controller): add ScanJob reconciler

fabriziosestito:
## Description

Adds the `ScanJob` reconciler.
When a `ScanJob` is created, if the target of the `ScanJob` Registry exists in the same namespace, a `CreateCatalog ' message is sent to the queue. 

Part of: #269

### DIFF
--- a/internal/controller/scanjob_controller.go
+++ b/internal/controller/scanjob_controller.go
@@ -1,0 +1,146 @@
+/*
+Copyright 2024.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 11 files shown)


internal/controller/scanjob_controller.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	// Check if resource is being deleted
+	if !scanJob.DeletionTimestamp.IsZero() {
+		return ctrl.Result{}, nil
+	}
+
+	if scanJob.IsFailed() || scanJob.IsComplete() || scanJob.IsInProgress() {
+		return ctrl.Result{}, nil
+	}
```
flavio, line 69: Right now this overlaps with `scanJob.IsInProgress()`, can we use this method? I think we're trying to achieve the same check, meaning that if a scanjob is already in progress the reconciler must exit

flavio (approved):
LGTM, I left a minor comment

internal/controller/scanjob_controller.go
flavio: you're right, I didn't read the code properly

internal/controller/scanjob_controller.go
```diff
@@ -0,0 +1,146 @@
...
+// +kubebuilder:rbac:groups=sbombastic.rancher.io,resources=scanjobs,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=sbombastic.rancher.io,resources=scanjobs/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=sbombastic.rancher.io,resources=scanjobs/finalizers,verbs=update
+
+// Reconcile reconciles a ScanJob object.
+func (r *ScanJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log := logf.FromContext(ctx)
+	log.Info("Reconciling ScanJob")
+
+	scanJob := &sbombasticv1alpha1.ScanJob{}
```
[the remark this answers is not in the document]
flavio: As discussed today:

- `MarkComplete`: is going to be called, but not as part of this PR. This is not going to be invoked by this reconciler
- `MarkFailed`: this is being called also inside of this PR

PR #280, opened 2025-06-24
chore: add ignores list to ignore the dependent bot PR to run commitlint

pohanhuangtw:
## Description
- Fix https://github.com/rancher-sandbox/sbombastic/pull/273

## Demo fork PR
- https://github.com/pohanhuangtw/sbombastic/pull/154


commitlint.config.js
```diff
@@ -1 +1,7 @@
-export default { extends: ['@commitlint/config-conventional'] };
+module.exports = {
+    extends: ['@commitlint/config-conventional'],
+    ignores: [(commit) => {
+      return commit.includes('Signed-off-by: dependabot[bot]');
```
fabriziosestito, line 4: I don't think this is the right fix. The actual error is:

`Signed-off-by: dependabot[bot] <support@github.com> ✖ body's lines must not be longer than 100 characters [body-max-line-length]` (see [action log](https://github.com/rancher-sandbox/sbombastic/actions/runs/15823715750/job/44598631218#step:7:24))

So, we might want to set the [body-max-line-length](https://commitlint.js.org/reference/rules.html#body-max-length) rule to `Infinity` instead.

fabriziosestito (approved):
LGTM 
Could you link the run on your fork that demonstrates that this works?

flavio (approved):

PR #281, opened 2025-06-24
feat: NATS deduplication

fabriziosestito:
### Description

Adds [message deduplication](https://docs.nats.io/using-nats/developer/develop_jetstream/model_deep_dive#message-deduplication) support to the `Publisher`.

This ensures that messages are idempotent within the default deduplication window of 2 minutes.
It is useful when the controller crashes or restarts before a scan job message is marked as "in progress", preventing duplicate processing when the message is resent.

### DIFF
--- a/internal/messaging/publisher.go
+++ b/internal/messaging/publisher.go
@@ -66,7 +66,7 @@ func (p *NatsPublisher) Publish(ctx context.Context, subject string, messageID s
 		Subject: subject,
 		Data:    message,
 		Header: nats.Header{
-			"Nats-Msg-Id": []string{messageID},
+			jetstream.MsgIDHeader: []string{messageID},
 		},
 	}
 	if _, err := p.js.PublishMsg(ctx, msg); err != nil {


pohanhuangtw (approved):
LGTM — just one question about de-duplication and retries.

JetStream keeps a Nats-Msg-Id in its duplicate-window cache for the whole window, even after a consumer reads and ACKs the message (see nats-io/nats-server #[2921](https://github.com/nats-io/nats-server/issues/2921)).

If an SBOM-generation worker ACKs the message but later fails, we would need to republish the task. Because the retry would carry the same MsgID, JetStream would flag it as a duplicate and drop it until the window expires.

Is my understanding correct?
If so, how do we want to handle retries?

1. Shorten the duplicate window?
2. Publish a new MsgID on each retry?
3. Delay the ACK until the worker has completed successfully?

I’d like to be sure we won’t end up with tasks that can’t be re-queued. Thanks!

flavio (approved):

PR #283, opened 2025-06-24
chore: change helm release name

pohanhuangtw:
## Description
- Fix release naming
- PR naming

## Demo from fork
- helm chart [PR](https://github.com/pohanhuangtw/sbombastic/pull/28) update from updatecli 
Screenshot 2025-06-25 at 2 33 18 PM

- [Release](https://github.com/pohanhuangtw/sbombastic/releases/tag/sbombastic-chart-0.1.0)
Screenshot 2025-06-25 at 2 33 24 PM

### DIFF
--- a/cr.yaml
+++ b/cr.yaml
@@ -1,0 +1,1 @@
+release-name-template: "{{.Name}}-chart-{{.Version}}"
--- a/.github/workflows/helm-chart-release.yml
+++ b/.github/workflows/helm-chart-release.yml
@@ -34,5 +34,7 @@ jobs:
 
       - name: Run chart-releaser
         uses: helm/chart-releaser-action@cae68fefc6b5f367a0275617c9f83181ba54714f # v1.7.0
+        with:
+          config: cr.yaml
         env:
           CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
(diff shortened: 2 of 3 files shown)


cr.yaml
```diff
@@ -0,0 +1 @@
+release-name-template: "{{.Name}}-chart-{{.Version}}"
```
holyspectral, line 1: Do you need the path to cr.yaml in [here](https://github.com/rancher-sandbox/sbombastic/blob/7f4c0e6ce969238e2f6b2a47287b85acd6bfcf7c/.github/workflows/helm-chart-release.yml#L36)?

For example,
```
  with:
    config: cr.yaml
```

holyspectral: Its default seems to be `$HOME/.cr.yaml` per [here](https://github.com/helm/chart-releaser?tab=readme-ov-file#common-usage).

pohanhuangtw: Thanks for the reviewing. 
I found that both` cr.yaml`, `.cr.yaml` works, however, I still want to follow the documentation.
Change to the `.cr.yaml` .

fabriziosestito: @pohanhuangtw, I am confused since the screenshot you attached still shows the wrong release name. As @holyspectral suggested, I think that we need to specify the path with an action input. For example, [here](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/1a4f99127bca2c02922321873088f54829d40fb6/.github/workflows/helm-chart-release.yaml#L46) the specified path is `.github/cr.yaml`

pohanhuangtw: I just test it out and found that as well, I am changing now.

PR #298, opened 2025-07-02
feat: support questions ui for rancher user.

pohanhuangtw:
## Description
- Fix https://github.com/rancher-sandbox/sbombastic/issues/236
- Support configure controller / worker / storage.

## Test (screen shot from rancher)
Screenshot 2025-07-02 at 2 58 15 PM
Screenshot 2025-07-02 at 2 58 22 PM
Screenshot 2025-07-02 at 2 58 30 PM

## TODO

- [ ] https://github.com/rancher-sandbox/sbombastic/issues/299

### DIFF
--- a/charts/sbombastic/questions.yaml
+++ b/charts/sbombastic/questions.yaml
@@ -1,0 +1,101 @@
+# This is a Rancher questions file
+---
+questions:
+  ###############################################################################
+  # Controller
+  ###############################################################################
+  - variable: controller.image.pullPolicy
+    label: Controller Image Pull Policy
+    type: enum
+    default: IfNotPresent
+    options:
+      - Always
+      - IfNotPresent
+      - Never
+    group: Controller
+    description: |
+      Pull policy of the Controller Deployment
+
+  - variable: controller.replicas
+    label: Controller Replicas
+    type: int
+    default: 3
+    min: 1
+    group: Controller
+    description: |
+      Number of replicas of the Controller Deployment
+
+  - variable: controller.logLevel
+    label: Controller Log Level
+    type: enum
+    default: info
+    options:
+      - debug
+      - info
+      - warn
+      - error
+    group: Controller
+    description: |
+      Log level of the Controller Deployment
+
+  ###############################################################################
+  # Worker
+  ###############################################################################
+  - variable: worker.image.pullPolicy
+    label: Worker Image Pull Policy
+    type: enum
+    default: IfNotPresent
+    options:
+      - Always
+      - IfNotPresent
+      - Never
+    group: Worker
+    description: |
+      Pull policy of the Worker Deployment
+
+  - variable: worker.replicas
+    label: Worker Replicas
+    type: int
+    default: 3
+    min: 1
+    group: Worker
+    description: |
+      Number of replicas of the Worker Deployment
+
+  - variable: worker.logLevel
+    label: Worker Log Level
+    type: enum
+    default: info
+    options:
+      - debug
+      - info
+      - warn
+      - error
+    group: Worker
+    description: |
+      Log level of the Worker Deployment
+
+  ###############################################################################
+  # Storage
+  ###############################################################################
+  - variable: storage.image.pullPolicy
+    label: Storage Image Pull Policy
+    type: enum
+    default: IfNotPresent
+    options:
+      - Always
+      - IfNotPresent
+      - Never
+    group: Storage
+    description: |
+      Pull policy of the Storage Deployment
+
+  - variable: storage.replicas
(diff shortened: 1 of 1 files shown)


charts/sbombastic/questions.yaml
```diff
@@ -0,0 +1,148 @@
...
+      - IfNotPresent
+      - Never
+    group: Storage
+    description: |
+      Pull policy of the Storage Deployment
+
+  - variable: storage.replicas
+    label: Storage Replicas
+    type: int
+    default: 1
```
fabriziosestito, line 144: when we will ship pgsql we can increase the replicas here, could you add a TODO comment and track this as part of the pgsql issue?

fabriziosestito (approved):

charts/sbombastic/questions.yaml
pohanhuangtw: Sure, I will add a TODO here, thanks.

pohanhuangtw:
Hi @fabriziosestito , just add the new line for the question.yaml.
Please take a look, thanks.

fabriziosestito (dismissed):

charts/sbombastic/questions.yaml
```diff
@@ -0,0 +1,149 @@
...
+    type: string
+    default: ghcr.io/rancher-sandbox/sbombastic/controller
+    group: Controller
+    description: |
+      Repository of the Controller Deployment
+
+  - variable: controller.image.tag
+    label: Controller Image Tag
+    type: string
+    default: v0.1.0-alpha1
```
flavio, line 18: This will become outdated as soon as we tag a new release.

I think we can remove this from the questions, I don't think we should let the user change the version of the stack so easily. I don't see the need for doing that.

@fabriziosestito, @pohanhuangtw: what do you think?

charts/sbombastic/questions.yaml
```diff
@@ -0,0 +1,149 @@
...
+    type: string
+    default: ghcr.io/rancher-sandbox/sbombastic/worker
+    group: Worker
+    description: |
+      Repository of the Worker Deployment
+
+  - variable: worker.image.tag
+    label: Worker Image Tag
+    type: string
+    default: v0.1.0-alpha1
```
flavio, line 71: Same here

charts/sbombastic/questions.yaml
```diff
@@ -0,0 +1,149 @@
...
+    type: string
+    default: ghcr.io/rancher-sandbox/sbombastic/storage
+    group: Storage
+    description: |
+      Repository of the Storage Deployment
+
+  - variable: storage.image.tag
+    label: Storage Image Tag
+    type: string
+    default: v0.1.0-alpha1
```
flavio, line 124: Same here

charts/sbombastic/questions.yaml
```diff
@@ -0,0 +1,149 @@
...
+  ###############################################################################
+  # Controller
+  ###############################################################################
+  - variable: controller.image.repository
+    label: Controller Image Repository
+    type: string
+    default: ghcr.io/rancher-sandbox/sbombastic/controller
+    group: Controller
+    description: |
+      Repository of the Controller Deployment
```
flavio, line 13: I double checked, Kubewarden doesn't offer that as part of our questions. That's because Rancher users set the registry mirror via a dedicated Rancher UI setting (which is global) that works via [`cattle.systemDefaultRegistry`](https://github.com/kubewarden/helm-charts/blob/cc056ebaa0b4e0425589f9a951dfd91c01a7cc46/charts/kubewarden-defaults/values.yaml#L62-L63).

Because of that, I would propose to remove this - and the other - image registry questions and file a bug to handle Rancher's global registry configuration

flavio (changes requested):

charts/sbombastic/questions.yaml
pohanhuangtw: You're right — I think we can just rely on the chart's appVersion to determine the image version, and there's no need to expose it as a question to the user.

I'll include this change as part of the bug fix that handles Rancher's global registry configuration.

charts/sbombastic/questions.yaml
pohanhuangtw: Rancher global registry [issue](https://github.com/rancher-sandbox/sbombastic/issues/309) created.

charts/sbombastic/questions.yaml
flavio: that's not what I meant, see the comments I left on the other issue. I meant the questions.yaml should not ask anything about the tag to use

flavio (approved):

PR #310, opened 2025-07-09
feat: support rancher global registry setting

pohanhuangtw:
## Description
- Fix https://github.com/rancher-sandbox/sbombastic/issues/309
- ensure the registry setting from global.cattle.systemDefaultRegistry only
- ensure  the tag use chart appVersion as default if tag in value is empty.
## Test

- make helm-unittest

### DIFF
--- a/charts/sbombastic/tests/controller/deployment_test.yaml
+++ b/charts/sbombastic/tests/controller/deployment_test.yaml
@@ -4,11 +4,14 @@ templates:
 tests:
   - it: "should render a Deployment with the correct replica count, logging level, image, and imagePullPolicy"
     set:
+      global:
+        cattle:
+          systemDefaultRegistry: rancher.io
       controller:
         replicas: 5
         logLevel: debug
         image:
-          repository: ghcr.io/rancher-sandbox/sbombastic/controller
+          repository: rancher-sandbox/sbombastic/controller
           tag: v0.1.0
           pullPolicy: Always
     asserts:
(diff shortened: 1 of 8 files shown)


charts/sbombastic/tests/controller/deployment_test.yaml
```diff
@@ -24,3 +24,17 @@ tests:
...
+          value: 3
+      - equal:
+          path: "spec.template.spec.containers[0].image"
+          value: "ghcr.io/rancher-sandbox/sbombastic/controller:0.1.0-alpha1"
+      - equal:
+          path: "spec.template.spec.containers[0].imagePullPolicy"
+          value: "IfNotPresent"
+      - contains:
+          path: "spec.template.spec.containers[0].args"
+          content: "-log-level=info"
```
flavio, line 40: I think we can remove this test, it would break in the future once we update the `AppVersion` because the tag would change.

I would rather put the assertion about `ghcr.io` being present inside of the previous test

charts/sbombastic/tests/storage/deployment_test.yaml
```diff
@@ -41,3 +41,14 @@ tests:
...
           path: "spec.template.spec.volumes[0].persistentVolumeClaim.claimName"
           value: "existing-claim"
+  - it: "should render a Deployment with the default values"
+    asserts:
+      - equal:
+          path: "spec.replicas"
+          value: 1
+      - equal:
+          path: "spec.template.spec.containers[0].image"
+          value: "ghcr.io/rancher-sandbox/sbombastic/storage:0.1.0-alpha1"
```
flavio, line 51: same here

charts/sbombastic/tests/worker/deployment_test.yaml
```diff
@@ -26,3 +26,17 @@ tests:
...
           path: "spec.template.spec.containers[0].args"
           content: "-log-level=debug"
+  - it: "should render a Deployment with the default values"
+    asserts:
+      - equal:
+          path: "spec.replicas"
+          value: 3
+      - equal:
+          path: "spec.template.spec.containers[0].image"
+          value: "ghcr.io/rancher-sandbox/sbombastic/worker:0.1.0-alpha1"
```
flavio, line 36: same here

fabriziosestito (approved):

PR #311, opened 2025-07-09
feat: add crd keep policy

pohanhuangtw:
## Description

- Fix https://github.com/rancher-sandbox/sbombastic/issues/243
- add helm.sh/resource-policy to keep the crd when uninstall
- remove image crd (registry CRD only)

## Test

- ensure make manifest will keep the `helm.sh\/resource-policy: keep`


charts/sbombastic/templates/crd/sbombastic.rancher.io_images.yaml
flavio, line 1: @fabriziosestito just to double check, we can remove this file since this Custom Resource is managed by `storage`. This is a leftover from the early days of the project. Isn't it?

fabriziosestito: yes it is

flavio (approved):

Issue #307, opened 2025-07-08
Implement error handling / retry and backoff for Nats messages


fabriziosestito:
Currently, if a message processing error occurs, the worker NAKs it, causing it to retry indefinitely. To prevent this, configure the stream with a backoff strategy and a maximum number of retries (see the Backoff setting here: https://docs.nats.io/nats-concepts/jetstream/consumers).

fabriziosestito:
Moving this outside the Scan Trigger epic #265

PR #328, opened 2025-07-21
refactor: ScanJob status conditions

fabriziosestito:
## Description

This PR improves the `ScanJob` status conditions by adding:

* **Scheduled**: set by the controller when the job is scheduled.
* **InProgress**: set by the workers when the job is running.
* **Pending helper**: a new helper to check if the job is still waiting to start.

These changes make it easier to see the exact status of a `ScanJob` without having to infer it from other status fields.

TBD: update the RFC

### DIFF
--- a/internal/controller/vulnerabilityreport_controller.go
+++ b/internal/controller/vulnerabilityreport_controller.go
@@ -84,6 +84,8 @@ func (r *VulnerabilityReportReconciler) Reconcile(ctx context.Context, req ctrl.
 	scanJob.Status.ScannedImagesCount = len(vulnerabilityReports.Items)
 	if scanJob.Status.ScannedImagesCount == scanJob.Status.ImagesCount {
 		scanJob.MarkComplete(sbombasticv1alpha1.ReasonAllImagesScanned, "All images scanned successfully")
+	} else {
+		scanJob.MarkInProgress(sbombasticv1alpha1.ReasonImageScanInProgress, "Image scan in progress")
 	}
 
 	if err := r.Status().Update(ctx, scanJob); err != nil {
--- a/internal/handlers/create_catalog.go
+++ b/internal/handlers/create_catalog.go
@@ -96,6 +96,15 @@ func (h *CreateCatalogHandler) Handle(ctx context.Context, message []byte) error
 	}
 	h.logger.DebugContext(ctx, "ScanJob found", "scanjob", scanJob.Name, "namespace", scanJob.Namespace)
 
+	if !scanJob.IsScheduled() {
+		h.logger.DebugContext(ctx, "ScanJob is not scheduled, skipping catalog creation", "scanjob", scanJob.Name, "namespace", scanJob.Namespace)
+		return nil
+	}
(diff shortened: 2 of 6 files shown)


flavio (approved):

internal/controller/vulnerabilityreport_controller.go
```diff
@@ -84,6 +84,8 @@ func (r *VulnerabilityReportReconciler) Reconcile(ctx context.Context, req ctrl.
 	scanJob.Status.ScannedImagesCount = len(vulnerabilityReports.Items)
 	if scanJob.Status.ScannedImagesCount == scanJob.Status.ImagesCount {
 		scanJob.MarkComplete(sbombasticv1alpha1.ReasonAllImagesScanned, "All images scanned successfully")
+	} else {
```
hhcs9527, line 87: If the scan fails during the middle of the operation (e.g., SBOM generation or scanning), would setting the status to InProgress still be accurate?

Currently, we always mark the job as InProgress when ScannedImagesCount < ImagesCount, but I think we should first check that the job has not already failed before updating the condition.

What do you think? Should we add a condition to prevent overwriting a Failed state?

internal/handlers/create_catalog.go
```diff
@@ -96,6 +96,15 @@ func (h *CreateCatalogHandler) Handle(ctx context.Context, message []byte) error
 	}
 	h.logger.DebugContext(ctx, "ScanJob found", "scanjob", scanJob.Name, "namespace", scanJob.Namespace)
 
+	if !scanJob.IsScheduled() {
+		h.logger.DebugContext(ctx, "ScanJob is not scheduled, skipping catalog creation", "scanjob", scanJob.Name, "namespace", scanJob.Namespace)
+		return nil
+	}
+	scanJob.MarkInProgress(v1alpha1.ReasonCatalogCreationInProgress, "Catalog creation in progress")
+	if err = h.k8sClient.Status().Update(ctx, scanJob); err != nil {
+		return fmt.Errorf("cannot update scan job status %s/%s: %w", createCatalogMessage.ScanJobNamespace, createCatalogMessage.ScanJobName, err)
```
hhcs9527, line 105: same here

hhcs9527 (commented):
Overall this looks good to me!
One suggestion: should we also check the sbom_generation handler for similar status update issues?

Also, I left a comment regarding concerns about resetting the status of a failed job — we might want to guard against that to avoid overwriting terminal states.

internal/controller/vulnerabilityreport_controller.go
```diff
@@ -84,6 +84,8 @@ func (r *VulnerabilityReportReconciler) Reconcile(ctx context.Context, req ctrl.
...
+		scanJob.MarkInProgress(sbombasticv1alpha1.ReasonImageScanInProgress, "Image scan in progress")
 	}
```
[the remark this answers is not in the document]
fabriziosestito: We don't have error handling yet; it will be introduced by #307. The logic will skip the operation if the ScanJob is in a final state.

alegrey91 (approved):
LGTM

PR #333, opened 2025-07-22
fix: ScanJob flow bugs

fabriziosestito:
## Description

This PR fixes some issues in the `ScanJob` flow:

* Removes the check for `ScanJob` being in the Scheduled state before handling. This avoids problems when the controller hasn’t updated the status yet by the time the message is received.
* Adds a retry on conflict when updating the status, to handle possible race conditions with the controller.
* Acks messages on error instead of nacking them, to avoid constant redelivery for now (this will be properly handled in #315).
* Temporarily disables deduplication for `ScanSBOM` messages until we add the `ScanJob` reference in #315 — we need the `ScanJob` UID to generate a proper deduplication ID.


internal/handlers/create_catalog.go
```diff
@@ -96,12 +97,20 @@ func (h *CreateCatalogHandler) Handle(ctx context.Context, message []byte) error
...
 	h.logger.DebugContext(ctx, "ScanJob found", "scanjob", scanJob.Name, "namespace", scanJob.Namespace)
 
-	if !scanJob.IsScheduled() {
-		h.logger.DebugContext(ctx, "ScanJob is not scheduled, skipping catalog creation", "scanjob", scanJob.Name, "namespace", scanJob.Namespace)
-		return nil
-	}
-	scanJob.MarkInProgress(v1alpha1.ReasonCatalogCreationInProgress, "Catalog creation in progress")
-	if err = h.k8sClient.Status().Update(ctx, scanJob); err != nil {
+	// It is possible that the controller is slow to set the status condition "Scheduled" to true,
+	// so we might encounter conflicts when settings the status condition to "InProgress".
```
flavio, line 101:
```suggestion
	// so we might encounter conflicts when setting the status condition to "InProgress".
```

internal/messaging/subscriber.go
```diff
@@ -63,7 +63,7 @@ func (s *NatsSubscriber) Run(ctx context.Context) error {
 					"error", err,
 				)
 
-				if err := msg.Nak(); err != nil {
+				if err := msg.Ack(); err != nil {
 					s.logger.ErrorContext(ctx, "Failed to nak message",
```
flavio, line 67: The error message has not been updated (it still references the Nak).

Can you also write a comment explaining why we are returning a Ack?

flavio (changes requested):
good findings, LGTM once the minor issues are addressed

internal/messaging/subscriber.go
fabriziosestito: The block is superfluous. I removed it as we are lacking after the error handling. 
NOTE: We will implement proper error handling in #315

pohanhuangtw (approved):

PR #358, opened 2025-08-01
feat: registry scan runner

### LINKED ISSUE #345: Implement periodic scanning
* Introduce the `period` field to the `Registry` CRD.
* Implement a [`Runnable`](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/manager#Runnable) that runs every X seconds, identifies registries requiring scans, and creates corresponding scan jobs.
* Ensure the use of a cached client.

fabriziosestito:
## Description

Implements the `Runnable` responsible for performing periodic scans based on the configured scan interval.

Closes #345


internal/controller/registry_scan_runner.go
```diff
@@ -0,0 +1,179 @@
...
+	if !lastScanJob.IsComplete() && !lastScanJob.IsFailed() {
+		log.V(1).Info("Registry has a running ScanJob, skipping.", "registry", registry.Name, "scanJob", lastScanJob)
+
+		return nil
+	}
+
+	if lastScanJob.Status.CompletionTime != nil {
+		timeSinceLastScan := time.Since(lastScanJob.Status.CompletionTime.Time)
+		if timeSinceLastScan < registry.Spec.ScanInterval.Duration {
+			log.V(2).Info("Registry doesn't need scanning yet", "registry", registry.Name, "timeSinceLastScan", timeSinceLastScan)
```
flavio, line 103: Nit: should we raise the log level of this message to keep it "hidden" with the default configuration? I wonder if we're not going to pollute the logs

flavio (approved):

pohanhuangtw (commented):
Hi @fabriziosestito ,

I have a design question regarding the registry runner.

Do you think it would make sense to re-create a ScanJob by updating a “last scanned” annotation on the corresponding Registry object? For example, we could have a goroutine that updates the annotation when a scan is triggered. This might make the periodic scan more precise compared to the current implementation.

That said, I think the current approach is simple enough and doesn’t require handling the extra complexity I just described — I just wanted to check.

As I understand it, the current implementation lists all registries every 1 minute. That means the actual scan time for a registry could be up to ~1 minute off from its intended interval. Is that expected?

Thanks for clarifying!

alegrey91 (approved):

PR #394, opened 2025-08-18
feat: support authentication for registries

### LINKED ISSUE #384: Implement native docker authentication for registries
Implement native Docker registry authentication using the `dockerconfigjson` format.

The `Registry` resource spec must include an `AuthSecret` containing credentials in `dockerconfigjson` format, which will be used by [Trivy](https://trivy.dev/v0.65/docs/advanced/private-registries/).

For reference on creating such secrets, see the [Kubernetes documentation](https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/#registry-secret-existing-credentials).

We need to create a `./docker/config.json` every time we launch trivy when generating SBOMs.

alegrey91:
## Description

Fix #384 

## Test


internal/handlers/create_catalog.go
```diff
@@ -121,6 +123,22 @@ func (h *CreateCatalogHandler) Handle(ctx context.Context, message []byte) error
...
+	if registry.IsPrivate() {
+		err = dockerauth.SetupDockerAuthForRegistry(ctx, h.k8sClient, registry)
+		if err != nil {
+			return fmt.Errorf("cannot setup docker auth: %w", err)
+		}
+		h.logger.DebugContext(ctx, "Setup registry authentication", "dockerconfig", os.Getenv("DOCKER_CONFIG"))
+		defer func() {
+			// uset the DOCKER_CONFIG variable so at every run
+			// we start from a clean environment.
+			if err = os.Unsetenv("DOCKER_CONFIG"); err != nil {
```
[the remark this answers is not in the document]
alegrey91: ok, agree

docs/private_registry.md
```diff
@@ -0,0 +1,62 @@
...
+metadata:
+  name: my-first-registry
+  namespace: default
+spec:
+  uri: dev-registry.default.svc.cluster.local:5000
+  scanInterval: 1h
+  authSecret: myregistrykey
+```
+
+This will allow SBOMbastic to scan for images from private registries.
```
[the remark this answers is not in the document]
alegrey91: Sure, thanks for pointing it out

charts/sbombastic/templates/worker/role.yaml
```diff
@@ -40,3 +40,9 @@ rules:
       - patch
       - update
       - watch
+  - apiGroups:
+      - ""
+    resources:
+      - secrets
+    verbs:
+      - get
```
pohanhuangtw, line 48: If we change it to this, and the E2E test is set up with a private registry, will the hardened Sbombastic deployment work as expected?

internal/handlers/dockerauth/auth.go
```diff
@@ -0,0 +1,79 @@
...
+	if authSecret.Type != corev1.SecretTypeDockerConfigJson {
+		return fmt.Errorf("secret is not of type %s", corev1.SecretTypeDockerConfigJson)
+	}
+	secretData := authSecret.Data[corev1.DockerConfigJsonKey]
+	dockerConfig, err := createDockerConfigJSON(registry.Spec.URI, secretData)
+	if err != nil {
+		return fmt.Errorf("cannot create dockerconfig file: %w", err)
+	}
+
+	err = os.Setenv("DOCKER_CONFIG", dockerConfig)
```
pohanhuangtw, line 41: As @flavio suggested, I think the generated temp docker configuration should be removed in this function

hack/private-registry.yaml
```diff
@@ -0,0 +1,84 @@
...
+kind: Pod
+metadata:
+  name: dev-registry
+  namespace: default
+  labels:
+    app: dev-registry
+spec:
+  containers:
+    - name: registry
+      image: registry:2.8.3
```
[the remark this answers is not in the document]
flavio: fine with me

flavio (dismissed):
Thanks for the having implemented all the requested changes

internal/handlers/dockerauth/auth.go
alegrey91: > The cleanup must be done by the callers of this function once they are done

Correct

api/v1alpha1/registry_types.go
```diff
@@ -47,6 +47,11 @@ type Registry struct {
 	Status RegistryStatus `json:"status,omitempty"`
 }
 
+// IsPrivate tells if the registry requires authentication
```
fabriziosestito, line 50:
```suggestion
// IsPrivate returns true when the registry requires authentication.
```

examples/private-registry/registry.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: sbombastic.rancher.io/v1alpha1
+kind: Registry
+metadata:
+  name: my-first-registry
+  namespace: default
+spec:
+  uri: dev-registry.default.svc.cluster.local:5000
+  scanInterval: 1h
+  authSecret: myregistrykey
```
fabriziosestito, line 9: could we call this secret something else? `my-auth-secret` for instance

examples/private-registry/secret.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: myregistrykey
```
fabriziosestito, line 4: could we call this secret something else? `my-auth-secret` for instance

internal/handlers/dockerauth/auth.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/docker/cli/cli/config/types"
+	"github.com/google/go-containerregistry/pkg/authn"
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/rancher/sbombastic/api/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+	k8stypes "k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// SetupDockerAuthForRegistry retrieve the Secret listed in the Registry resource
```
fabriziosestito, line 20: maybe `BuildDockerConfigForRegistry` ?

internal/handlers/create_catalog.go
```diff
@@ -121,6 +123,26 @@ func (h *CreateCatalogHandler) Handle(ctx context.Context, message []byte) error
 		return fmt.Errorf("cannot create transport for registry %s: %w", registry.Name, err)
 	}
 	registryClient := h.registryClientFactory(transport)
+	// if authSecret value is set, then setup Docker
+	// authentication to get access to the registry
+	if registry.IsPrivate() {
```
fabriziosestito, line 128: This part of the handler is not tested. Let's add a test for it.

internal/handlers/utils_test.go
```diff
@@ -0,0 +1,137 @@
...
+	authUser     = "user"
+	authPass     = "password"
+	registryURI  = "ghcr.io/rancher-sandbox/sbombastic/test-assets"
+	imageName    = "golang"
+	tag          = "1.12-alpine"
+	platform     = "linux/amd64"
+	digest       = "sha256:1782cafde43390b032f960c0fad3def745fac18994ced169003cb56e9a93c028"
+)
+
+type RegistryTestSuite struct {
```
fabriziosestito, line 29: The current name is ambiguous. Considering the method names, it could easily be confused with the `suite` from [testify](https://pkg.go.dev/github.com/stretchr/testify/suite). Would it be clearer to rename this  `testPrivateRegistry` (no need to be public)?

internal/handlers/utils_test.go
```diff
@@ -0,0 +1,137 @@
...
+	registry    *registry.RegistryContainer
+	registryURL string
+	ctx         context.Context
```
fabriziosestito, line 32: Storing ctx in a struct field is an antipattern and it should be avoided, see: https://go.dev/blog/context-and-structs

internal/handlers/utils_test.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+func setupPrivateRegistry(t *testing.T) *RegistryTestSuite {
```
fabriziosestito, line 35: This actually starts a registry. Could we name it  `startTestPrivateRegistry`?

fabriziosestito, line 35: there is no need to accept testing.T, we could just return error from here and let the caller deal with thee assertions/require

fabriziosestito: Just accept a logger as a constructor parameter and store the slog logger inside the struct field

internal/handlers/utils_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+	err = dockerCli.ImageTag(ctx, image, ref)
+	if err != nil {
+		return fmt.Errorf("failed to tag image %s: %w", image, err)
+	}
+
+	return nil
+}
+
+func (suite *RegistryTestSuite) tearDown(t *testing.T) {
```
fabriziosestito, line 131: let's call this `stop`. it should accept ctx.Context and return error so we can remove testing.T

Tiltfile
```diff
@@ -5,7 +5,13 @@ update_settings(k8s_upsert_timeout_secs=300)
 
 # Setup a development registry so we can push images to it
 # and use them to test the scanner.
-k8s_yaml("./hack/registry.yaml")
+# To setup a private registry, set `use_private_registry: true`
+# on the configuration file.
+use_private_registry = settings.get("use_private_registry", False)
```
fabriziosestito, line 10: could you update the tiltfile settings example file, adding this setting commented out?

internal/handlers/utils_test.go
```diff
@@ -0,0 +1,137 @@
...
+	registry    *registry.RegistryContainer
+	registryURL string
+	ctx         context.Context
+}
+
+func setupPrivateRegistry(t *testing.T) *RegistryTestSuite {
+	registryContainer, err := registry.Run(
+		t.Context(),
+		registry.DefaultImage,
+		registry.WithHtpasswd("user:$2y$10$nTQigvLRGGHCBQwZB4MPPe2SA6GYG218uTe1ntHusNcEjLaAfBive"), // user:password
```
fabriziosestito, line 39: nit: let's use a constant

fabriziosestito (approved):
:shipit:

flavio (approved):

PR #416, opened 2025-09-02
fix(api): registry scanInterval optional

fabriziosestito:
## Description

This PR makes `scanInterval` optional; leaving it unset disables automatic scanning.

### DIFF
--- a/api/v1alpha1/registry_types.go
+++ b/api/v1alpha1/registry_types.go
@@ -13,7 +13,9 @@ type RegistrySpec struct {
 	Repositories []string `json:"repositories,omitempty"`
 	// AuthSecret is the name of the secret in the same namespace that contains the credentials to access the registry.
 	AuthSecret string `json:"authSecret,omitempty"`
-	// ScanInterval is the interval at which the registry is scanned. Set to 0 to disable automatic scanning.
+	// ScanInterval is the interval at which the registry is scanned.
+	// If not set, automatic scanning is disabled.
+	// +kubebuilder:validation:Optional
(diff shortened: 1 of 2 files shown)


api/v1alpha1/registry_types.go
```diff
@@ -13,7 +13,9 @@ type RegistrySpec struct {
 	Repositories []string `json:"repositories,omitempty"`
 	// AuthSecret is the name of the secret in the same namespace that contains the credentials to access the registry.
 	AuthSecret string `json:"authSecret,omitempty"`
-	// ScanInterval is the interval at which the registry is scanned. Set to 0 to disable automatic scanning.
+	// ScanInterval is the interval at which the registry is scanned.
+	// If not set, automatic scanning is disabled.
+	// +kubebuilder:validation:Optional
 	ScanInterval metav1.Duration `json:"scanInterval"`
```
alegrey91, line 19: Shouldn't we define it as a pointer instead?

```golang
ScanInterval *metav1.Duration `json:"scanInterval"`
```

So that will not be shown when we do `kubectl get registry` if it's not set.

fabriziosestito: This is commonly used without a pointer in k8s: 
https://github.com/kubernetes/kubernetes/blob/cb163eb55b83ac54e6af10d9c8b7faaa36d4b141/pkg/proxy/apis/config/types.go#L74

alegrey91:
@fabriziosestito should we update the RFC too?

flavio (approved):

PR #426, opened 2025-09-04
chore: Automate Golang version update

olblak:
## Description

To automate the Golang version update, this pull request introduces the following changes

### Unset Golang version from GHA workflow

To reduce the number of Golang version, we need to update after new
Golang version, I configure the GitHub action workflow to use
the version defined in the project go.mod
More information on:
  * https://github.com/actions/setup-go?tab=readme-ov-file#getting-go-version-from-the-gomod-file

## Check Golang version update

Add an Updatecli manifest to monitor the Golang version defined in the Dockerfile and the go.mod
Add a new GHA workflow to run Updatecli every hour to check for update.
I used the Updatecli compose file to make it easier in the future to add more automation

## refactor the Helm chart automate update

Simplify the GHA workflow used for updating helm chart which rely on Updatecli to use an Updatecli compose file
to make it easier in the future to add more update pipeline

## Test

Right now tested manually

### DIFF
--- a/.github/workflows/updatecli.yaml
+++ b/.github/workflows/updatecli.yaml
@@ -1,0 +1,28 @@
+name: Updatecli
+on:
+  workflow_dispatch:
+  schedule:
+    # Run every hour
+    - cron: "0 * * * *"
+jobs:
+  updatecli:
+    name: Run Updatecli
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write # for updatecli to update the repository
+      pull-requests: write # for updatecli to create a PR
+    steps:
+      - name: Checkout
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+
+      - name: Install Updatecli in the runner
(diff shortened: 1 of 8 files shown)


.github/workflows/updatecli.yaml
```diff
@@ -0,0 +1,28 @@
+name: Updatecli
+on:
+  workflow_dispatch:
+  schedule:
+    # Run every hour
+    - cron: "0 * * * *"
```
hhcs9527, line 6: Hi @olblak ,
I’m wondering why the update is scheduled every hour instead of being triggered on merge.

If the goal is to unify the version, wouldn’t it make more sense to update it only when a bump commit is merged? That way, the version stays in sync with the actual changes and avoids unnecessary updates.

Could you clarify the reasoning behind this schedule?

Thanks!

pohanhuangtw (dismissed):

flavio (approved):

flavio:
@olblak can you fix the commit messages to make the linter happy?

fabriziosestito (dismissed):

PR #440, opened 2025-09-09
chore: bump trivy-version

pohanhuangtw:
## Description

- Bumps Trivy to v0.66.0.
- Bumps k8s.io/apiextensions-apiserver to v0.34.0 to resolve an OpenTelemetry/GRPC interceptor symbol mismatch that surfaced with older k8s.io/* deps.
- Implements minimal methods required by storage.Interface (Stats, CompactRevision, SetKeysFunc) so the custom store compiles against apiserver 0.34.x.

## Test


pohanhuangtw:
Hi @flavio @fabriziosestito 
I'd appreciate it if you could review PR.
This PR bumps the Trivy dependency, which initially caused CI failures due to a version conflict with k8s.io/apiserver. I've now aligned all k8s.io/* module versions to resolve the issue.

pohanhuangtw:
Hi @flavio ,
Thanks for the feedback. The trivy bump involve many k8s related bump, that is why we have the PR.

After checking with @fabriziosestito , we found that e2e test failed because of GC sync failure was actually caused by the new WatchList behavior introduced in k8s v0.34. Our storage layer doesn’t support it yet.

### Next steps:

- We’ll keep the Trivy-only bump scoped separately.
- The k8s group bump (with the WatchList workaround) will be handled in its own PR.
- In the meantime, Fabrizio disabled WatchList in our tests as a temporary workaround until we can implement proper support.

Big thanks to Fabrizio for helping track down the root cause 🙏

PR #443, opened 2025-09-09
feat: allow user to customize resources

flavio:
Allow the user to tune the resources (CPU, memory) that SBOMbastic pods are allowed to use.

Currently the limits of the controller are too low, when a scanjob happens the reconciler can exceed the amount of memory the controller is entitled to use. As a result, the controller is terminated by the OOM killer.

Warning: let's discuss what kind of limits and requests we should put as default values for the different components of the stack.

### DIFF
--- a/charts/sbombastic/templates/controller/deployment.yaml
+++ b/charts/sbombastic/templates/controller/deployment.yaml
@@ -46,13 +46,10 @@ spec:
               port: 8081
             initialDelaySeconds: 5
             periodSeconds: 10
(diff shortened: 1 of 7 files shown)


fabriziosestito (approved):

flavio:
I've raised the memory limits of the storage to 3Gb. I've a scanned 345 images and it uses almost 1Gb of memory.

The memory usage of the storage component will probably decrease significantly once we move away from sqlite.

fabriziosestito (approved):

PR #445, opened 2025-09-10
fix(handlers): use info level for regular logging instead of debug

fabriziosestito:
## Description

use info level for regular logging instead of debug in the handlers


flavio (approved):
thanks, I really needed this change

PR #464, opened 2025-09-15
refactor: switch to pgsql pt1

### LINKED ISSUE #289: helm: update components required by sqlite
Now that sqlite is not being used by storage anymore, we can simplify our helm chart and remove all the resources and configurations related with that (like the PVC).

## Acceptance criteria

- All the resources required by sqlite are removed from the helm chart

fabriziosestito:
## Description

First PR to switch from SQLite and sqlx to PostgreSQL using pgx.
This change also starts a CNPG cluster in the Helm chart. 

Remaining tasks, including making the CNPG installation optional and configuring TLS connections, will be done in a follow-up PR.

We also decided to move away from [squirrel](https://github.com/Masterminds/squirrel) since it is unmaintained. 

fix #289

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -233,6 +233,10 @@ require (
 	github.com/huandu/xstrings v1.5.0 // indirect
 	github.com/in-toto/in-toto-golang v0.9.0 // indirect
 	github.com/inconshreveable/mousetrap v1.1.0 // indirect
+	github.com/jackc/pgpassfile v1.0.0 // indirect
+	github.com/jackc/pgservicefile v0.0.0-20240606120523-5a60cdf6a761 // indirect
+	github.com/jackc/pgx/v5 v5.7.6 // indirect
+	github.com/jackc/puddle/v2 v2.2.2 // indirect
 	github.com/jbenet/go-context v0.0.0-20150711004518-d14ea06fba99 // indirect
 	github.com/jedisct1/go-minisign v0.0.0-20230811132847-661be99b8267 // indirect
 	github.com/jmespath/go-jmespath v0.4.1-0.20220621161143-b0104c826a24 // indirect
@@ -357,6 +361,7 @@ require (
 	github.com/subosito/gotenv v1.6.0 // indirect
 	github.com/syndtr/goleveldb v1.0.1-0.20220721030215-126854af5e6d // indirect
 	github.com/tchap/go-patricia/v2 v2.3.3 // indirect
+	github.com/testcontainers/testcontainers-go/modules/postgres v0.39.0 // indirect
 	github.com/tetratelabs/wazero v1.9.0 // indirect
 	github.com/theupdateframework/go-tuf v0.7.0 // indirect
 	github.com/titanous/rocacheck v0.0.0-20171023193734-afe73141d399 // indirect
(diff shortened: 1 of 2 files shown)


flavio (changes requested):
great job. Overall it looks good to me. I would like to be less aggressive with the inline docs (meaning: preserve it)

Tiltfile
```diff
@@ -52,6 +52,20 @@ helm_resource(
...
+    "cnpg/cloudnative-pg",
+    namespace="cnpg-system",
+    flags=[
+        "--create-namespace",
+    ],
+    resource_deps=[
+        "cnpg-repo",
+    ],
+)
+
```
pohanhuangtw, line 68: CNPG (CloudNativePG) instance may take some time to become fully ready. Would this delay cause any issues for SBOMbastic when attempting to use the database before CNPG is fully initialized?

Would it make sense to explicitly declare a dependency on the CNPG resource to avoid potential race conditions?
For example, by adding the following to line 79 of the file:

```bash
resource_deps = [
    "cnpg-repo",
],
```
what do you think?

pohanhuangtw (commented):
Overall great work!
Left some concern about the start up delay for cnpg to be fully ready.

Tiltfile
fabriziosestito: > Would this delay cause any issues for SBOMbastic when attempting to use the database before CNPG is fully initialized?

This scenario will be handled within the storage code, since it can also occur in production.

In this context, the goal is only to verify that the CRDs are installed correctly before continuing with the Helm chart installation.
That verification is already performed in steps 108 and 109.

charts/sbombastic/templates/storage/deployment.yaml
```diff
@@ -35,13 +35,14 @@ spec:
 {{ toYaml .Values.storage.resources | indent 12 }}
           {{- end }}
           volumeMounts:
-            - name: sqlite-pvc
-              mountPath: /data/sqlite/
             - name: cert-volume
               mountPath: /certs
+            - name: cnpg-secret
```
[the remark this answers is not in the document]
fabriziosestito: great idea, this means we do not have to switch this off if the user does not want to use cnpg

cmd/storage/main.go
```diff
@@ -36,20 +19,38 @@ func main() {
...
+		os.Exit(1)
+	}
+	if _, err := db.Exec(ctx, storage.CreateSBOMTableSQL); err != nil {
+		logger.Error("failed to create sbom table", "error", err)
+		os.Exit(1)
+	}
+	if _, err := db.Exec(ctx, storage.CreateVulnerabilityReportTableSQL); err != nil {
+		logger.Error("failed to create vulnerability report table", "error", err)
+		os.Exit(1)
+	}
```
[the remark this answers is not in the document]
fabriziosestito: yes but let me take care of this in the next PR

internal/storage/store_test.go
```diff
@@ -28,14 +30,54 @@ const keyPrefix = "/storage.sbombastic.rancher.io/sboms"
...
+	ctx := context.Background()
+
+	pgContainer, err := postgres.Run(ctx,
+		"postgres:16-alpine",
+		postgres.WithDatabase("testdb"),
+		postgres.WithUsername("test"),
+		postgres.WithPassword("test"),
+		testcontainers.WithWaitStrategy(
+			wait.ForLog("database system is ready to accept connections").
+				WithOccurrence(2).
```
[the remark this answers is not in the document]
fabriziosestito: No, this means we are waiting for this log twice. 
I found a better way to wait for the container to be ready, so I am changing this.

cmd/storage/main.go
```diff
@@ -36,20 +19,38 @@ func main() {
...
-		log.Fatalln(err)
+		logger.Error("failed to read database URI", "error", err)
+		os.Exit(1)
 	}
 
-	db.SetMaxOpenConns(1) // Avoid connection pool issues
-	db.MustExec(storage.CreateImageTableSQL)
-	db.MustExec(storage.CreateSBOMTableSQL)
-	db.MustExec(storage.CreateVulnerabilityReportTableSQL)
+	db, err := pgxpool.New(ctx, string(dbURI))
```
pohanhuangtw, line 30: Just to confirm — is the case where CNPG is not fully ready handled here?

fabriziosestito: Yes, we will need to introduce backoff in the connection. I am investigating how.

flavio (approved):

PR #475, opened 2025-09-23
feat: postgres TLS support

fabriziosestito:
## Description

This change introduces TLS connection support for PostgreSQL.  
Currently, all parameters required to connect with CNPG are hardcoded. In upcoming PRs, these parameters will be made configurable via the Helm chart, and will be exposed via env variables till we don't fix https://github.com/rancher-sandbox/sbombastic/issues/476.

By performing initialization within the `BeforeConnect` callback, certificates are reloaded correctly whenever a connection is created or reset. This ensures that updated certificates on disk are automatically applied ([discussion](https://github.com/jackc/pgx/discussions/2103)).

Additionally, TLS authentication for replica clients is now supported.  

For reference, see the [CNPG cluster example with cert-manager](https://cloudnative-pg.io/documentation/current/samples/cluster-example-cert-manager.yaml).

### DIFF
--- a/cmd/storage/main.go
+++ b/cmd/storage/main.go
@@ -1,12 +1,18 @@
 package main
 
 import (
+	"context"
+	"crypto/tls"
+	"crypto/x509"
+	"errors"
+	"fmt"
 	"log/slog"
 	"os"
 
 	genericapiserver "k8s.io/apiserver/pkg/server"
 	"k8s.io/component-base/cli"
 
+	"github.com/jackc/pgx/v5"
 	"github.com/jackc/pgx/v5/pgxpool"
 
 	"github.com/rancher/sbombastic/cmd/storage/server"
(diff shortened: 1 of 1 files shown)


alegrey91 (dismissed):
LGTM

cmd/storage/main.go
```diff
@@ -31,7 +37,40 @@ func run() int {
...
+			RootCAs:            caCertPool,
+			ServerName:         config.ConnConfig.Host,
+			MinVersion:         tls.VersionTLS12,
+			InsecureSkipVerify: false,
+		}
+
+		return nil
+	}
+
+	db, err := pgxpool.NewWithConfig(ctx, config)
```
pohanhuangtw, line 73: Is there any chance the database might not be ready before establishing the connection here?

pohanhuangtw (commented):
Overall looks good to me 👍
Is the current implementation only supporting mTLS for CNPG?

cmd/storage/main.go
flavio: According to [the docs](https://pkg.go.dev/github.com/jackc/pgx/v5/pgxpool#pkg-overview) this is not checking for the connection to be actually established:

> A pool returns without waiting for any connections to be established. Acquire a connection immediately after creating the pool to check if a connection can successfully be established. 

This looks like something we should add below to make the init process a bit more reliable, isn't it?

fabriziosestito: @flavio @pohanhuangtw I’ll track this as debt for after the epic.

It’s a bit trickier than it looks: we’ll need readiness and liveness probes in the storage component, using the connection pool’s `Ping` to check availability before taking requests.

I’d also like to move migrations into an init command (run in an init container), but that needs #476 first.

Let me dig into this a bit more before opening an issue. We can chat about it in the next planning. For now, just letting Kubernetes restart the pod works fine for the scope of this epic.

fabriziosestito: @pohanhuangtw
> Is the current implementation only supporting mTLS for CNPG?

We’re not planning to add mTLS (x509 client certificate auth) for Postgres right now.

In Postgres, mTLS means setting up client certificate authentication in `pg_hba.conf` with the `cert` method (see [[docs](https://www.postgresql.org/docs/current/ssl-tcp.html#SSL-CLIENT-CERTIFICATES)](https://www.postgresql.org/docs/current/ssl-tcp.html#SSL-CLIENT-CERTIFICATES)). Just sending a cert is not enough because proper user mapping is also required. CNPG handles this with a kubectl plugin, but for our MVP we’re sticking with password auth (via a secret), which is sufficient for now since the CNPG operator generates a random user/password combination for each cluster.

If it turns out people really need cert-based auth, we can add support later.

**Note:** mTLS authentication is currently used only for replicas to authenticate with each other.

For now, we only require TLS with server certificate verification.

PR #478, opened 2025-09-24
Feat/pg helm configuration

### LINKED ISSUE #287: helm: integrate PostgreSQL
This is a first step to make our helm chart aware of PostgreSQL.

This is the user experience we want to achieve:

- By default, we take care of creating a PostgreSQL that is used by SBOMbastic
- The user can disable the creation of the database and just point us to an existing one to be used

## Acceptance criteria

- [x] By default a PostgreSQL database is created, these are its requirements:
  - Must be persistent
  - Credentials must be randomly created and stored inside of a Secret
  - Connection must be secured with a certificate that is generated by cert-manager
  - Certificate rotation must be handled
- [x] The user can disable the creation of the PostgreSQL database and point it to an existing one. The user must also point us to a Secret that contains the user credentials. It also must provide us a ~ConfigMap~ Secret that holds the CA used to validate the TLS endpoint
- [x] All the data (Secrets, ConfigMaps) are mounted into the storage pods
- [x] Ensure everything works smoothly inside the tilt dev environment
- [x] The CNPG instances and storage (class and size) are tunable
- [x] The user can opt-out from the CNPG cluster creation and provide their own secret

fabriziosestito:
## Description

Fix #287


charts/sbombastic/values.yaml
```diff
@@ -36,6 +36,32 @@ storage:
     requests:
       cpu: 250m
       memory: 300Mi
+  postgres:
+    # Name of existing secret containing the URI to connect to external Postgres.
+    # Expects a `uri` key. The sslmode parameter will be ignored.
```
flavio, line 41: I would add an example of how the `uri` should look like or suggest to read our docs (I haven't checked the docs PR, hopefully there's something that covers that in there)

flavio (dismissed):

PR #480, opened 2025-09-25
test: add test for VEXHub application on scans

### LINKED ISSUE #469: test: add test for VEXHub application on scans
We currently lack a test verifying that a VEXHub is correctly applied to a scan. Specifically, the test should confirm that scan results include suppressed vulnerabilities when a matching VEX entry exists.

To address this, we need to:

- Extend the scan job handler tests to cover this scenario, creating another expected report that contains suppressed vulnerabilities.
- Add a fake VEXHub containing a VEX file that applies to the test image:
  - Create a VEXHub containing just a vex file that matches the test image used in the test, see https://github.com/aquasecurity/vex-repo-spec
  - The VEXHub structure includes a `main.tar.gz` file and other content, but only `main.tar.gz` is required for the test.
  - Reference the tar gz in the `.well-known/vex-repository.json` repository.
  - Remove all the other files in the `test/fixtures/vexhub` folder

**Expected outcome:**
When the VEXHub repository is enabled, the scan report should include suppressed vulnerabilities as defined by the VEX file.

alegrey91:
## Description

Fix #469 

## Test

```
go test -v ./internal/handlers/... -run TestScanSBOMHandler_Handle$
```


internal/handlers/generate_sbom.go
```diff
@@ -186,8 +186,8 @@ func (h *GenerateSBOMHandler) generateSBOM(ctx context.Context, image *storagev1
 		"--disable-telemetry",
 		"--cache-dir", h.workDir,
 		"--format", "spdx-json",
-		"--db-repository", "public.ecr.aws/aquasecurity/trivy-db",
-		"--java-db-repository", "public.ecr.aws/aquasecurity/trivy-java-db",
+		"--skip-db-update",
+		"--skip-java-db-update",
```
flavio, line 190: How are we going to control the update flow of the vulnerability dbs after this change gets merged?

flavio: good catch!

flavio (approved):

PR #495, opened 2025-10-01
feat(storage): TLS and auto-reload of certificates

### LINKED ISSUE #241: Add TLS certificates to all components
To improve our system's security posture, we should ensure that all internal components communicate over encrypted channels using TLS and, where possible, enforce mutual TLS (mTLS) for identity verification.

We will use [cert-manager](https://cert-manager.io/) to automate certificate management.

* [ ] Configure mTLS for NATS to secure messaging between components (nodes, clients)
* [ ] Enable TLS for the Storage API server extension to protect data access
* [ ] Enable TLS for the controller’s webhook to ensure secure admission control

AC:
- Components should be able to auto-reload certificates when the certs are refreshed without downtime.

fabriziosestito:
## Description
This PR adds TLS certificate generation via cert-manager to the storage component.

It also configures automatic certificate reloading using the apiserver's [`DynamicCertKeyPairContent`](https://pkg.go.dev/k8s.io/apiserver/pkg/server/dynamiccertificates) utility, which uses fsnotify under the hood to watch for certificate file changes.

Fixes #241

### DIFF
--- a/charts/sbombastic/templates/storage/apiservice.yaml
+++ b/charts/sbombastic/templates/storage/apiservice.yaml
@@ -2,11 +2,12 @@ apiVersion: apiregistration.k8s.io/v1
 kind: APIService
 metadata:
   name: v1alpha1.storage.sbombastic.rancher.io
+  annotations:
+    cert-manager.io/inject-ca-from: {{ .Release.Namespace }}/{{ include "sbombastic.fullname" . }}-storage-tls
   labels:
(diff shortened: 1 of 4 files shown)


flavio (approved):
Nice finding (I'm referring to the apiserver library that does exactly what we need)

alegrey91 (approved):
LGTM

PR #500, opened 2025-10-03
fix(worker): add variant to platform

fabriziosestito:
## Description
Fixes the platform string to include the variant field when discovering multi-arch images.

Previously, the variant (e.g., `v7` in `linux/arm/v7`) was omitted from the platform string, which could cause incorrect image selection for ARM variants.


flavio (approved):
LGTM, debugged and fixed that in pair session with @fabriziosestito

PR #503, opened 2025-10-07
feat(messaging): implement NATS retry with backoff

fabriziosestito:
## Description

Fix: #307 

Implements NATS subscriber retry and backoff using `NakWIthDelay`

### DIFF
--- a/internal/messaging/subscriber.go
+++ b/internal/messaging/subscriber.go
@@ -4,10 +4,27 @@ import (
 	"context"
 	"fmt"
 	"log/slog"
+	"math"
+	"math/rand/v2"
+	"time"
 
 	"github.com/nats-io/nats.go"
 	"github.com/nats-io/nats.go/jetstream"
 )
+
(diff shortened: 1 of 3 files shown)


internal/messaging/subscriber.go
```diff
@@ -37,6 +62,9 @@ func NewNatsSubscriber(ctx context.Context, nc *nats.Conn, durable string, handl
 		jetstream.ConsumerConfig{
 			FilterSubjects: subjects,
 			Durable:        durable,
+			AckWait:        1 * time.Hour,
```
alegrey91, line 65: Just to clarify, could you let me know why we need this?

fabriziosestito: I will add a comment to explaing what this is for

fabriziosestito:
Moving back to in progress, we can improve this by using the `InProgress` acknowledgement in slow handlers

PR #505, opened 2025-10-07
chore(helm): add trivy database arguments

### LINKED ISSUE #412: Provide configuration for the vulnerability database repository via the Helm chart
Building on #385 and #411, 
We need to add a Helm chart value to configure the vulnerability database repository. 

This is particularly useful in air-gapped environments.

AC:
- [x] Add a value to the `values.yaml` file to set `trivy-vulnerability-db` and `trivy-java-vulnerability-db` in the worker component
- [x] Update `questions.yaml` file

alegrey91:
## Description

Fix #412 

## Test


charts/sbombastic/templates/worker/deployment.yaml
```diff
@@ -28,6 +28,12 @@ spec:
           args:
             - -nats-url
             - {{ .Release.Name }}-nats.{{ .Release.Namespace }}.svc.cluster.local:4222
+            {{- if .Values.worker.trivyDBRepository }}
+            - -trivy-db-repository={{ .Values.worker.trivyDBRepository }}
```
flavio, line 32: I think it would be safer to [quote](https://helm.sh/docs/chart_template_guide/functions_and_pipelines/) the input given by the user here

charts/sbombastic/templates/worker/deployment.yaml
```diff
@@ -28,6 +28,12 @@ spec:
...
+            {{- end }}
+            {{- if .Values.worker.trivyJavaDBRepository }}
+            - -trivy-java-db-repository={{ .Values.worker.trivyJavaDBRepository }}
```
flavio, line 35: same here

charts/sbombastic/values.yaml
```diff
@@ -83,6 +83,8 @@ worker:
     requests:
       cpu: 250m
       memory: 300Mi
+  trivyDBRepository: public.ecr.aws/aquasecurity/trivy-db
+  trivyJavaDBRepository: public.ecr.aws/aquasecurity/trivy-java-db
```
flavio, line 87: Is there no need to specify the protocol (http/https)?

charts/sbombastic/questions.yaml
```diff
@@ -75,6 +75,22 @@ questions:
     description: |
       Log level of the Worker Deployment
 
+  - variable: worker.trivyDBRepository
+    label: Worker Trivy DB Repository
+    type: string
+    default: public.ecr.aws/aquasecurity/trivy-db
+    group: Worker
+    description: |
+      OCI repository to retrieve trivy-db
```
flavio, line 84: Are we sure this is an OCI repository? AFAIK it's also possible to pull this data from an S3 bucket

charts/sbombastic/values.yaml
flavio: Sorry, my fault. They have their own container registry. That's totally fine then. Ignore my comment.

flavio (dismissed):

fabriziosestito (changes requested):
Could you add a Helm unit test testcase in the `charts/sbombastic/tests/worker` ?

"it should render the vulnerability database flags if the flags are set"

PR #538, opened 2025-10-17
docs: update issue template

flavio:
Mention the troubleshooting guidelines when opening an issue.

I **hope** the link works as expected when the issue is rendered.

Update, it works. I've tested that on a personal repo:

image


flavio:
I've updated the docs adding a warning section and fixing the link to the debug script.

olblak:
Would it make sense to add check mark

---

- [ ] I reviewed the logs and confirmed they contain no sensitive information.

---

Maybe with a link that explain the kind of data the script is collecting

flavio:
> I would make it clear that we collect those information, but I would postpone the decision to re-work this until it becomes a real issue for issue reporter. Meanwhile, people can always reach on different channels

The script is already printing on the STDOUT the names of the types of resources being collected

olblak (approved):

PR #542, opened 2025-10-17
fix: typo in quickstart guide for ScanJob

juadk:
## Description

This PR fixes just a small issue in the scan job section.


fabriziosestito (changes requested):
Thanks and sorry for the nitpick

PR #555, opened 2025-10-22
feat: registry platform filter

### LINKED ISSUE #513: feat: registry platform filter
Currently we scan all the versions of a multi-architecture image, but this leads to longer scanning time and higher memory usage.

The end goal is to avoid scanning images of platform and architectures that are not relevant to the user.

## Proposal

Change the `Registry` spec by adding this new field:

```yaml
spec:
  platforms: # list of platforms that are relevant, empty by default
    - arc: amd64 # mandatory
      os: linux # mandatory
    - arc: arm
      os: linux
      variant: v7 # optional
```

## Acceptance Criteria

- [x] Update the `Registry` to allow platform filtering
- [ ] If `Registry.spec.platforms` has values, we will generate `Image`, `SBOM` and `VulnerabilityReport` only for the images that belong to the platforms provided by the user
- [ ] Ensure that resources generated are updated after a ScanJob is run. For example, assume the user defines a Registry without specifying `spec.platforms` and immediately scans it. After the scan is done all the platforms have been scanned. Next, add a filter by specifying a plaftorm. After the next scan is done oly the relevant platforms should be kept.

alegrey91:
## Description

Fix #513 

## Test


fabriziosestito (dismissed):

api/v1alpha1/registry_types.go
```diff
@@ -29,6 +31,9 @@ type RegistrySpec struct {
 	CABundle string `json:"caBundle,omitempty"`
 	// Insecure allows insecure connections to the registry when set to true.
 	Insecure bool `json:"insecure,omitempty"`
+	// Platforms allows to specify the list of platform to scan.
+	// If not set, the catalog will be triggered to list all the available platforms on the registry.
```
flavio: Given this is going to be shown to end users, I would be a bit more generic:

```suggestion
	// If not set, all the available platforms of a container image will be scanned.
```

internal/webhook/v1alpha1/platforms.go
```diff
@@ -0,0 +1,64 @@
...
+		return fmt.Errorf("unsupported arch %s for OS %s", p.Arch, p.OS)
+	}
+
+	// Check variant
+	variants, hasVariants := allowedVariants[p.Arch]
+	if hasVariants {
+		// This arch supports variants, check if provided variant is valid
+		if p.Variant == "" {
+			return fmt.Errorf("arch %s requires a variant (one of: %v)", p.Arch, variants)
+		}
```
flavio: This is something that concerns me. For example, in this file we state that `arm64` can have only the `v8` variant.

If you look at the multi-arch image manifest of the latest worker image:

```console
$ crane manifest ghcr.io/kubewarden/sbomscanner/worker:v0.7.1
...
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:0dc61c6fbf356064a66e7cfda953b9f88d321cda8ebbafc1fbd0f147319857b0",
      "size": 2749,
      "platform": {
        "architecture": "arm64",
        "os": "linux"
      }
    },
...
```

As you can see here, the `v8` variant is omitted.

I fear that a user would be forced by this validation to specify linux arm64 as "linux" + "arm64" + "v8", which would result in a `linux/arm64/v8` string that will fail this code from the `isPlatformAllowed` function:

```golang
return slices.ContainsFunc(allowedPlatforms, func(p v1alpha1.Platform) bool {
		return p.String() == platform.String()
	})
```

flavio (changes requested):
Overall LGTM. I left some minor comments, but I also have a doubt related with the business logic

internal/handlers/create_catalog.go
```diff
@@ -566,3 +570,23 @@ func computeImageUID(ref name.Reference, digest string) string {
...
+
+	// If no platform is specified in the Registry CR,
+	// we assume the user wants to scan all the platforms.
+	if len(allowedPlatforms) == 0 {
+		return true
+	}
+
+	return slices.ContainsFunc(allowedPlatforms, func(p v1alpha1.Platform) bool {
+		return p.String() == platform.String()
+	})
```
flavio: This has to be changed, I think it would not work (as in consider the plaform not allowed) these cases:

- user specified `linux/arm64/v8`, but the manifest has `linux/arm64`
- user specified `linux/arm` because he wants all the arm 32 bit variants to be scanned, but the manifest has `linux/arm/v[6,7,8]`

flavio (approved):
LGTM

fabriziosestito (approved):

PR #562, opened 2025-10-23
feat: add init containers and probes

### LINKED ISSUE #483: feat(worker): add graceful NATS connection handling and health probes
We need to gracefully handle the NATS connection when NATS isn't available yet. Instead of exiting, the worker should retry connecting. We should look into NATS's built-in reconnection settings for this.

We also need to implement health probes:
- [ ] liveness probe (checks NATS connection is healthy)
- [ ] startup probe (gives time for initial NATS connection)
- [ ] readiness probe (marks pod ready once NATS is connected)
- [ ] configure probes in the Helm chart deployment

All three probes could use the same endpoint.

### LINKED ISSUE #484: feat(controller): add graceful startup when API extension storage types and NATS  are unavailable
The controller currently exits with errors if the API extension server isn't up yet and the types it watches aren't available, or if NATS isn't reachable. Since these aren't CRDs (they're served by the API extension server), they won't be registered until that server is ready. This causes crashloop backoff during initial deployments.
Instead, the controller should:

 - [ ] retry watching API extension server types with backoff instead of exiting
 - [ ] retry NATS connection with backoff instead of exiting
 - [ ] log warnings during retry attempts for observability
 - [ ] only mark ready once it successfully establishes watches and NATS connection

### LINKED ISSUE #485: feat(storage): add graceful PostgreSQL startup and migration init container
The storage API extension server currently exits with errors if PostgreSQL isn't available on startup. This causes crashloop backoff during initial deployments.
Instead, we should:

The storage API extension server needs two improvements for PostgreSQL handling:
1. Migration init container

 - [x] add init container that runs migrations before the main container starts
- [x] use the same component binary with an --init flag(this depends on #476)
- [x]  implement retry with backoff if PostgreSQL isn't available yet
- [x]  configure init container in the Helm chart deployment

2. Health probes

- [x]  implement liveness/readiness probes using [AddLivezChecks](https://pkg.go.dev/k8s.io/apiserver/pkg/server#GenericAPIServer.AddLivezChecks)
- [x]  use [pgxpool.Ping](https://pkg.go.dev/github.com/jackc/pgx/v4/pgxpool#Pool.Ping) for PostgreSQL health checks
 - [x] configure probes in the Helm chart deployment

fabriziosestito:
## Description

Fixes #485
Fixes #484
Fixes #483

This update introduces the following changes:

* **Init containers** have been added to all components:

  * **Controller, Worker:** wait until the storage types are available and the JetStream cluster is reachable before starting.
  * **Storage:** wait for PostgreSQL to become available, then perform database migrations.
* **Health probes** (liveness and readiness) have been added to the **Storage** and **Worker** components.

### DIFF
--- a/cmd/worker/main.go
+++ b/cmd/worker/main.go
@@ -72,6 +72,7 @@ func main() {
 		cancel()
 	}()
 
+	config := ctrl.GetConfigOrDie()
 	natsOpts := []nats.Option{
 		nats.RootCAs(natsCAFile),
 		nats.ClientCert(natsCertFile, natsKeyFile),
@@ -80,8 +81,12 @@ func main() {
 	if init {
 		logger = logger.With("task", "init")
 
-		err = cmdutil.WaitForJetStream(ctx, natsURL, natsOpts, logger)
-		if err != nil {
+		if err := cmdutil.WaitForStorageTypes(ctx, config, logger); err != nil {
+			logger.Error("Error waiting for storage types", "error", err)
+			os.Exit(1)
+		}
+
+		if err := cmdutil.WaitForJetStream(ctx, natsURL, natsOpts, logger); err != nil {
 			logger.Error("Error waiting for JetStream", "error", err)
 			os.Exit(1)
 		}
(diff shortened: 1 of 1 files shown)


fabriziosestito:
@flavio @alegrey91 moving this back to In Progress. I’ll be incorporating the logic from the other components into the same PR, as I realized that some parts of the code are shared.

alegrey91:
Can we consider it ready for review?

PR #571, opened 2025-10-28
docs(rfc): add documentation for scan filter

### LINKED ISSUE #612: RFC describing the problem and proposed solution
Write an RFC describing the implementation approach for #460.

### LINKED ISSUE #645: Write image tag filters RFC
As the title says, we need to document the new image tag filters with an RFC.

alegrey91:
## Description

Fix #612  
Fix #645

## Test

### DIFF
--- a/docs/rfc/0005_scan_filter.md
+++ b/docs/rfc/0005_scan_filter.md
@@ -1,0 +1,122 @@
+|              |                                 |
+| :----------- | :------------------------------ |
+| Feature Name | Scan Filters                    |
+| Start Date   | Oct 27th, 2025                  |
+| Category     | Architecture                    |
+| RFC PR       | [#555](https://github.com/kubewarden/sbomscanner/pull/555), [#614](https://github.com/kubewarden/sbomscanner/issues/614) |
+| State        | **ACCEPTED**                    |
+
+# Summary
+
+[summary]: #summary
+
+Define a way to filter the images being scanned.
+
+# Motivation
+
+[motivation]: #motivation
+
+The purpose of this RFC is to define a way to filter certain parameters in order to speed up the scan time.
+Currently, when defining a Registry, the only way to limit the contents being scanned is to provide a list of repositories. However, all the images found inside these repositories are being scanned. This can lead to scanning a lot of irrelevant images, wasting a lot of time and resources.
+
+We should provide a way to filter the images being scanned.
+
+We would start focusing on two kinds of filters:
+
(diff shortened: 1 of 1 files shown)


docs/rfc/0005_scan_filter.md
```diff
@@ -0,0 +1,113 @@
...
+## Examples / User Stories
+
+[examples]: #examples
+
+- As a user, I want to scan only `linux/amd64` based images on my registry.
+- As a user, I want to scan `linux/amd64` and `linux/arm` based images on my registry.
+- As a user, I want to scan only the most recent tags of images on my registry.
+- As a user, I want to scan only the most recent tags, `linux/amd64` based images on my registry.
+- As a user, I want to scan only for non release candidate tags.
+- As a user, I want to scan for images deployed in a production environment with tags that end in `-prod`.
```
fabriziosestito: Those stories look good, but I'd add a generic one to cover the overall filtering capability:

- As a user, I want to filter image tags using regular expressions, string comparisons, or semantic version operations.

docs/rfc/0005_scan_filter.md
```diff
@@ -0,0 +1,113 @@
...
+      variant: "v7"
+```
+
+A `platforms` field not set means that no filters are going to be applied to the scan. This will result in a complete scan with image catalogization.
+
+## Tags
+
+The tags filter will be used to avoid scanning unwanted tags for the given image. This will help the user focus only on the tags they really want to scan. We will leverage the CEL expression language to filter for image tags, enabling support for the semver library.
+
+We will provide the user with a new field on the `Registry` CRD called `matchConditions`, under the `repositories` field. For this feature, we are going to introduce a breaking change, since `repositories` is currently a `string`, and we are going to replace it with a new struct to host the `name` of the repository and the `matchConditions` struct. The `matchConditions` field will contain an arbitrary list of CEL expressions to define with high precision the list of tags to scan.
```
fabriziosestito: We're missing:
- Link to CEL Kubernetes documentation
- Which operations we support (string, regex, and semver from Kubernetes)

Add these links:
- CEL introduction: https://kubernetes.io/docs/reference/using-api/cel/
- CEL language repository: https://github.com/google/cel-go
- CEL spec repository: https://github.com/google/cel-spec
- String functions: https://github.com/google/cel-spec/blob/master/doc/langdef.md#string-functions
- Semver operations: https://kubernetes.io/docs/reference/using-api/cel/#kubernetes-semver-library

fabriziosestito (changes requested):

PR #573, opened 2025-10-29
ci: sign multi arch image index manifest

flavio:
Prior to this commit, we did NOT sign the container images built by our pipelines. We only signed the artifacts attached to the final GitHub release (like the SBOM and the provenance file).

With this commit we sign (using both cosign v2 and v3 formats):

- architecture specific images
- multi-arc image manifest

Notes to reviewer: I had to rename one file changing its suffix. That makes reviewing the whole PR a bit noisy. I suggest you to look at the two commits that are part of the PR, especially the 1st one.

You can see the whole pipeline run [here](https://github.com/flavio/sbomscanner/actions/runs/18911787488), on my personal fork. Note that the "create-release" step failed because I didn't have a draft GitHub release on my personal fork. The failure can be ignored, since the changes have been done to the previous steps.
Also, the image can now be verified properly:

```console
$ cosign verify \
  --certificate-identity-regexp="https://github.com/flavio/sbomscanner/.*" \
  --certificate-oidc-issuer="https://token.actions.githubusercontent.com" \
  --offline=true \
  --use-signed-timestamps=true \
  ghcr.io/flavio/sbomscanner/controller:v0.7.2-rc1

Verification for ghcr.io/flavio/sbomscanner/controller:v0.7.2-rc1 --
The following checks were performed on each of these signatures:
  - The cosign claims were validated
  - Existence of the claims in the transparency log was verified offline
  - The code-signing certificate was verified using trusted certificate authority certificates

[{"critical":{"identity":{"docker-reference":"ghcr.io/flavio/sbomscanner/controller:v0.7.2-rc1"},"image":{"docker-manifest-digest":"sha256:89120d2d46b6fc6d52512cb7c9c137e870c86eafb3ac6db73f181b30c9daf9fa"},"type":"https://sigstore.dev/cosign/sign/v1"},"optional":null}]
```


flavio:
JFYI: [this](https://github.com/flavio/sbomscanner/actions/runs/18945162278) is a run with the updated GHA on my personal fork. As said before, ignore the failure, to us what matters is that the merge step worked fine.

viccuad (approved):
LGTM! Looking forward to reusing the approach for the other components.

For SLSA v3, the same OIDC token needs to be used for the build and release. The OIDC token will have metadata of the workflow file path and run. Sadly, one cannot use reusable workflows, as they cannot be pinned by digest, only by tag, and tags are mutable.

For the admission controller we solved this by repeating ourselves between the container-build.yml: (for main and feature branches)
https://github.com/kubewarden/policy-server/blob/f7bf225b325cfb717cb95771bd3334b4ed8a1618/.github/workflows/container-build.yml#L13-L81
and the release.yml: (for tags)
https://github.com/kubewarden/policy-server/blob/main/.github/workflows/release.yml#L16-L87

It may be worth it to copy-paste the steps of `container-build.yml` into `release.yml` to achieve SLSA v3 and verification via `slsactl`.

jvanz (approved):

flavio:
@viccuad thanks for the comment. I think we can address that in a future PR.

PR #576, opened 2025-10-30
fix(controller): add ReconciliationTimeout as global option

fabriziosestito:
## Description

Adds a reconciliation timeout to prevent reconcilers from hanging indefinitely.
Thanks to @bigkevmcd for highlighting this issue.

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -26,21 +26,22 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/nats-io/nats.go"
 
-	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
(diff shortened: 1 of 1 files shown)


cmd/controller/main.go
```diff
@@ -204,6 +206,10 @@ func main() {
 				},
 			},
 		},
+		Controller: config.Controller{
+			// ReconciliationTimeout is used as the timeout passed to the context of each Reconcile call.
+			ReconciliationTimeout: 30 * time.Second,
```
flavio: Does that change mean that all our reconciler will timeout after 30 seconds? I think it's a good idea to have an explicit timeout, but maybe 30 seconds is too low as a default value

fabriziosestito: bumped to 90 seconds as agreed

PR #603, opened 2025-11-06
docs(rfc): multiscan process

### LINKED ISSUE #583: RFC: write multiscan RFC

alegrey91:
## Description

Fix #583 

## Test

### DIFF
--- a/docs/rfc/0006_multiscan.md
+++ b/docs/rfc/0006_multiscan.md
@@ -1,0 +1,97 @@
+|              |                                 |
+| :----------- | :------------------------------ |
+| Feature Name | Multi-Scan                      |
+| Start Date   | 6 November 2025                 |
+| Category     | Architecture                    |
+| RFC PR       | [fill this in after opening PR] |
+| State        | **ACCEPTED**                    |
+
+# Summary
+
+[summary]: #summary
+
+Support multiple scanning tools.
+
+# Motivation
+
+[motivation]: #motivation
+
+We want to add support for multiple scanning tools (such as grype) in order to enrich the vulnerability reports, making it more complete and accurate.
+
+This will also allow us to be less vendor centric, since we are currently relying only on trivy to generate SBOMs and scan for vulnerabilities.
+
+## Examples / User Stories
+
+[examples]: #examples
+
+### User story 1
+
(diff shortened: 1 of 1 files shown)


docs/rfc/0006_multiscan.md
```diff
@@ -0,0 +1,128 @@
...
+
+```yaml
+apiVersion: sbomscanner.kubewarden.io/v1alpha1
+kind: ScanJob
+metadata:
+  name: scanjob-example
+  namespace: default
+spec:
+  registry: example-registry
+  multiScan: true
```
flavio: Note: that poses a question about which scanner is the default one when `multiScan` is disabled.

docs/rfc/0006_multiscan.md
```diff
@@ -0,0 +1,128 @@
...
+
+# Motivation
+
+[motivation]: #motivation
+
+We want to add support for `grype` in order to enrich the vulnerability reports, making them more complete and accurate.
+
+This will allow us to be vendor-neutral, since we are currently relying only on `trivy` to generate SBOMs and scan for vulnerabilities.
+
+Additionally, we discovered that `grype` is able to find more vulnerabilities than `trivy`. Below is a recap of our research:
```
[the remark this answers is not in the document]
alegrey91: absolutely. that's a good point.

docs/rfc/0006_multiscan.md
flavio: yes, but let's clarify that inside of the RFC

fabriziosestito: I am not sure whether this setting belongs at the Registry level or at the Installation level, or if we should even add it in the first iteration.

I also don’t think a `multiScan` field is explicit enough.
It would be clearer to let users select the engines, for example:
```
scanEngines: [trivy, grype]
```

fabriziosestito (changes requested):

fabriziosestito:
Closing this as not planned for now; we can revisit and continue this work when we’re ready.

flavio:
> Closing this as not planned for now; we can revisit and continue this work when we’re ready.

Shouldn't we merge this PR but with status REJECTED?

PR #623, opened 2025-11-12
test(handlers): refactor handlers test to use a real registry and remove mocked registry client

### LINKED ISSUE #456: test: refactor CreateCatalogHandler tests to use a real registry.
Right now, the `CreateCatalogHandler` tests use a mocked registry client. That makes things painful because we have to mock every single request for different cases.

- [ ] Let’s switch the tests to use a real registry instead (either our `ghcr.io` registry or a testcontainer one works).
- [ ] Also check the private registry test, as it looks like it’s still mocked, so we’re not really testing auth at the moment.
- [ ] add a single arch case to the create catalog handle test

fabriziosestito:
## Description

Fix #456

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -492,6 +492,7 @@ linters:
           - gosec
           - noctx
           - wrapcheck
+          - gocognit
       - path: '_test\.go'
         linters:
           - revive


alegrey91 (approved):
lgtm

flavio (approved):

Issue #460, opened 2025-09-12
EPIC: provide a way to filter the images being scanned


flavio:
## Description

Right now, when a **Registry** is defined, the only way to limit what gets scanned is by listing specific repositories. The problem is that once those repos are discovered, **all** images inside them get scanned, even the ones we don’t care about. This can waste a lot of time and resources scanning irrelevant images.

We should introduce a way to **filter images before scanning**.

We’ll start with two kinds of filters:

* **Platform filtering:**
  Many repositories publish multi-architecture images. For example, a single image might have variants for `amd64`, `arm64`, `s390x`, and more. Most users only care about the architecture that actually runs in their cluster — e.g., `amd64` for x86_64 nodes. We should allow filtering so that only relevant platforms are scanned.

* **Tag filtering (using CEL):**
  Some repositories keep tons of old or irrelevant tags. It would be great to let users define which tags matter using a [CEL expression](https://github.com/google/cel-spec). That way, they can easily match tags using regex, semantic version checks, or other custom logic.

###  User Stories

* **As a user**, I want to filter images by platform (e.g., `linux/amd64`) so that only images compatible with my cluster are scanned, avoiding unnecessary scans for architectures I don’t use.

* **As a user**, I want to filter images by tag using CEL expressions so that I can control which images get scanned, for example, by including only certain versions or matching specific tag patterns.

PR #634, opened 2025-11-17
feat: add CEL tag evaluator

fabriziosestito:
## Description

Introduce a CEL-based tag evaluator that filters tags using CEL expressions, extended with Kubernetes’ [semver CEL library](https://kubernetes.io/docs/reference/using-api/cel/#kubernetes-semver-library).

Note: Regular expressions are built into CEL and can be applied using `<string>.matches`.

Part of: #460


alegrey91 (approved):
LGTM

PR #636, opened 2025-11-17
feat(api)!: add matchConditions field to registry

### LINKED ISSUE #635: Change the Registry CRD so that it is possible to specify a CEL expression for each repository defined.
Add a matchConditions field to each repository in the Registry CRD to filter tags using CEL expressions.
This is similar to what k8s does to match requests in webhook configurations or ValidationAdmissionPolicies
See: 
- https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-matchconditions
- https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/#matching-requests-matchconditions

```yaml
apiVersion: sbomscanner.kubewarden.io/v1alpha1
kind: Registry
metadata:
  name: my-first-registry
  namespace: default
spec:
  uri: dev-registry.default.svc.cluster.local:5000
  repositories:
    - name: "nginx"
      matchConditions:
        - name: "every tag except the latest"
          expression: "tag != 'latest'"
  platforms:
    - arch: "amd64"
      os: "linux"
```

alegrey91:
## Description

This PR introduces support for image tag filters.
The user is now capable of filtering by image tag specifying one or more `matchCondition`s for each `Repository` in the registry.
Closes #635 

## Test

### DIFF
--- a/internal/controller/registry_controller.go
+++ b/internal/controller/registry_controller.go
@@ -72,11 +72,10 @@ func (r *RegistryReconciler) reconcileRegistry(ctx context.Context, registry *v1
 		return ctrl.Result{}, fmt.Errorf("unable to list Images: %w", err)
 	}
 
-	var repositories []string
+	allowedRepositories := sets.New[string]()
 	for _, repo := range registry.Spec.Repositories {
-		repositories = append(repositories, repo.Name)
+		allowedRepositories.Insert(repo.Name)
 	}
-	allowedRepositories := sets.NewString(repositories...)
 
 	tagEvaluator, err := cel.NewTagEvaluator()
 	if err != nil {


api/v1alpha1/registry_types.go
```diff
@@ -50,6 +50,25 @@ type RegistryStatus struct {
 	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,1,rep,name=conditions"`
 }
 
+// Repository contains fields to identify repositories from the catalog.
+type Repository struct {
+	// Name is an identifier for the image name to be scanned.
+	Name string `json:"name"`
+	// MatchConditions is a list of CEL expressions to identify the image tag.
```
fabriziosestito:
```suggestion
	// MatchConditions filters image tags using CEL expressions.
```

internal/filters/tag_test.go
```diff
@@ -0,0 +1,116 @@
...
+		{
+			name:         "matches single condition",
+			tagEvaluator: te,
+			matchConditions: []v1alpha1.MatchCondition{
+				{
+					Name:       "no images with -dev tags",
+					Expression: "!tag.matches('$-dev')",
+				},
+			},
+			tag:     imageTag,
```
fabriziosestito: nitpick: consider using the literal string value instead of the variable to make the test more explicit. That way readers don't need to jump to the variable definition to see what's being tested.

Issue #628, opened 2025-11-14
Bug: Watch events not propagated across storage replicas


fabriziosestito:
The current storage watch implementation uses an in-memory broadcaster that's isolated per replica. When running multiple replicas of the storage backend, watch clients connected to replica A won't receive events for mutations happening on replica B.

### Current behavior

1. Client connects to replica A, calls `Watch()` → gets watch stream from replica A's broadcaster
2. Another client hits replica B, performs a `Create()`/`Update()`/`Delete()`
3. Replica B triggers `s.broadcaster.Action()` on its own in-memory broadcaster
4. Watch client on replica A receives nothing

Each replica has its own `watch.Broadcaster` instance that only knows about local watchers. Mutations call `broadcaster.Action()` locally but this doesn't propagate to other replicas.

### Expected behavior

Watch events should be received by all watchers across all replicas, regardless of which replica handles the mutation.

### Impact

- Multi-replica deployments don't work correctly for watch
- Kubernetes controllers connected to one replica miss events from other replicas 
- The garbage collector controller is affected by the same problem, breaking owner reference cleanup (See #622)
- Unpredictable behavior during rolling updates or pod restarts

### Proposed solution

Use PostgreSQL LISTEN/NOTIFY to propagate events across replicas:
- Mutations send NOTIFY within the transaction
- Each replica runs a LISTEN goroutine
- Received notifications are broadcast locally to connected watchers

PR #642, opened 2025-11-20
feat(storage): distributed watch

fabriziosestito:
## Description

Fixes: #628 

This PR introduces a NATS-based broadcaster and watcher so that watch events are propagated across replicas. It uses NATS Core rather than JetStream, as persistence is not required. The storage API server has been updated to run all watcher goroutines within the same errgroup, instead of running them as fire-and-forget routines. This ensures that a failure in any watcher causes the entire group to stop as intended.

## Test

Store tests have been updated to use the new broadcaster and watcher.

### Trade-off

Because only references to objects are sent, preventing large vulnerability reports or SBOMs from filling the NATS queue, we cannot retrieve a deleted object from the store to publish it as a watch event. 

Instead, a dummy object is constructed using only the metadata of the deleted object. It does not include other fields such as the vulnerability report or SBOM data. Watchers, therefore, do not receive the full object when deletion events occur. This is sufficient for the current reconciler implementation and can be revisited if soft-delete is implemented in the future.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,6 +9,7 @@ require (
 	github.com/aws/smithy-go v1.23.2
 	github.com/docker/cli v29.0.4+incompatible
 	github.com/go-logr/logr v1.4.3
+	github.com/google/cel-go v0.26.1
 	github.com/google/go-cmp v0.7.0
 	github.com/google/go-containerregistry v0.20.7
 	github.com/google/uuid v1.6.0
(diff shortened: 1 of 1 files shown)


alegrey91:
@fabriziosestito can you please fix the golangci-lint error?

flavio (dismissed):
wow, what a PR! Thanks a lot, I think it's well done and can be merged

alegrey91 (dismissed):

PR #647, opened 2025-11-21
docs: update user-guide with image tag filters

### LINKED ISSUE #644: Write user-guide for image tag filter
Write user-guide for the new feature, giving the user some tutorial on how to use the `matchConditions` to select a subset of images on the registry.

alegrey91:
## Description

This PR adds user-guide section about image tag filtering.
Fix #644 

## Test

### DIFF
--- a/examples/registry.yaml
+++ b/examples/registry.yaml
@@ -7,6 +7,11 @@ spec:
   uri: dev-registry.default.svc.cluster.local:5000
   catalogType: OCIDistribution
   scanInterval: 1h
+  repositories:
+    - name: nginx
+      matchConditions:
+        - name: tags greater than v1.12.0
+          expression: "semver(tag, true).isGreaterThan(semver('v1.12.0', true))"
   platforms:
     - arch: "amd64"
       os: "linux"


docs/user-guide/scanning-registries.md
```diff
@@ -152,7 +153,61 @@ spec:
       variant: "v7"
 ```
 
-## 5. Monitor Scan Progress
+## 5. Filtering By Tags
+
+Another way to reduce the scanning load is to apply filters to the image tags. To use this filter, you have to use the `matchConditions` field, provided under `repositories`.
+
+This field allows the user to write their own [CEL](https://kubernetes.io/docs/reference/using-api/cel/) expressions to filter for image tags. Let's see an example:
```
fabriziosestito:
```suggestion
You can filter image tags using CEL](https://kubernetes.io/docs/reference/using-api/cel/) expressions in the `matchConditions` field under `repositories`. This lets you use regex patterns, semantic versioning comparisons, or string operations (like substring matching or contains checks) to select which tags to scan. This filtering capability helps reduce the number of scanned images and lowers the load on the system.

Let's see an example:
```

I rephrased the paragraph to focus more on the feature rather than the load of the system

examples/registry.yaml
```diff
@@ -7,6 +7,11 @@ spec:
   uri: dev-registry.default.svc.cluster.local:5000
   catalogType: OCIDistribution
   scanInterval: 1h
+  repositories:
+    - name: image/repository/name
```
fabriziosestito: let's use a realistic example, such as `nginx`

examples/registry.yaml
```diff
@@ -7,6 +7,11 @@ spec:
...
+      matchConditions:
+        - name: tags greater than v1.2.0
```
fabriziosestito: if we go with nginx, we can use > 1.12.0 (ingress controller CVE fixed)

PR #657, opened 2025-11-27
feat(storage): pagination

### LINKED ISSUE #52: Implement pagination and fix resourceVersion in storage.GetList method
We want to implement pagination in the `GetList` method.

When the user sends `opts.Predicate.Limit` true we will need to generate a continue token and limit the number of items we return.
If the user sends a continue token, we will need to parse it and return to the next page until there is nothing more to return.

Use https://github.com/kubernetes/apiserver/blob/master/pkg/storage/etcd3/store.go as a reference implementation.

Moreover, we will need to implement the list `resourceVersion`
For instance, grafana uses snowflake IDs https://github.com/intergral/grafana/blob/2f390b310c1eec9d9cb7ee8459090d5dc19d3c02/pkg/apiserver/storage/file/file.go#L92

fabriziosestito:
## Description

Adds cursor-based pagination to the storage layer using an auto-increment `id` column. The continue token is just the last ID as a string, and queries use `WHERE id > cursor ORDER BY id` to fetch the next page.

We fetch `limit + 1` rows to determine if there's a next page, which avoids returning an empty final page when using label or field selectors.

## Notes

The `id` column is not the primary key:`(namespace, name)` still is, so this doesn't break anything. 
We can reconsider making `id` the primary key when we break the schema for deduplication.

Fixes #52

### DIFF
--- a/internal/apiserver/storage.go
+++ b/internal/apiserver/storage.go
@@ -16,7 +16,6 @@ import (
 	"k8s.io/apiserver/pkg/registry/rest"
 	genericapiserver "k8s.io/apiserver/pkg/server"
 	"k8s.io/apiserver/pkg/server/dynamiccertificates"
-	"k8s.io/apiserver/pkg/server/healthz"
 	genericoptions "k8s.io/apiserver/pkg/server/options"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	basecompatibility "k8s.io/component-base/compatibility"
@@ -114,7 +113,6 @@ func NewStorageAPIServer(db *pgxpool.Pool, nc *nats.Conn, certFile, keyFile stri
 
 	databaseChecker := newDatabaseChecker(db, logger)
 	serverConfig.AddReadyzChecks(databaseChecker)
-	serverConfig.AddHealthChecks(healthz.PingHealthz)
 
 	// Create generic server
 	genericServer, err := serverConfig.Complete().New("sbom-storage-apiserver", genericapiserver.NewEmptyDelegate())


internal/apiserver/storage.go
```diff
@@ -114,7 +113,6 @@ func NewStorageAPIServer(db *pgxpool.Pool, nc *nats.Conn, certFile, keyFile stri
 
 	databaseChecker := newDatabaseChecker(db, logger)
 	serverConfig.AddReadyzChecks(databaseChecker)
-	serverConfig.AddHealthChecks(healthz.PingHealthz)
```
alegrey91: Do you have any specific reason to remove this health check?

fabriziosestito: As indicated in the commit message, this is already included by default. Adding it a second time may introduce race conditions that lead to storage failures (depending on the goroutine start timing). This was causing the e2e tests to fail sometimes, introducing flakiness.

internal/storage/store.go
```diff
@@ -359,9 +386,18 @@ func (s *store) GetList(ctx context.Context, key string, opts storage.ListOption
...
 	}
 
+	var lastID int64
+	var count int64
+	var hasMore bool
 	for rows.Next() {
+		if opts.Predicate.Limit > 0 && count == opts.Predicate.Limit {
+			hasMore = true
+			break
+		}
```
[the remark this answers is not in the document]
fabriziosestito: Yes. This simulates Kubernetes behaviour. When selectors are used, there is no way to know in advance whether the next page contains items. The only reliable method is to fetch one additional page: if that extra page contains items, a continue token is returned. This prevents returning a continue token that would otherwise lead to an empty page when used.

flavio (approved):

PR #665, opened 2025-12-01
fix: proper invocation of helm schema

flavio:
Change `Makefile` to run helm schema using `go run`. This ensure it works properly also on machines where the binary is missing.

Moreover, use the proper flag to specify the `values.yaml` file.


flavio:
Superseded by https://github.com/kubewarden/sbomscanner/pull/696

PR #694, opened 2025-12-12
feat(cel): add cost estimation and limits

fabriziosestito:
## Description

Adds cost estimation and limits to the CEL tag evaluator to prevent overly complex expressions from slowing down evaluation or DoS-ing the controller.

kubernetes uses a similar approach, see: https://github.com/kubernetes/kubernetes/issues/126819

Also, it limits the repository matchconditions to 10.


flavio (approved):
good catch!

alegrey91 (approved):

PR #696, opened 2025-12-12
fix: helm schema invocation

alegrey91:
## Description

Originally provided by @flavio in this PR: https://github.com/kubewarden/sbomscanner/pull/665
Change Makefile to run helm schema using `go run`. This ensure it works properly also on machines where the binary is missing.

Moreover, use the proper flag to specify the `values.yaml` file.


fabriziosestito:
@flavio there are two different projects, the latter is the one we use in the ci

PR #703, opened 2025-12-15
perf(controller): watch only VulnerabilityReport metadata instead of the full object

fabriziosestito:
## Description

Switch VulnerabilityReport controller to metadata-only watch. We only need labels and deletion timestamp, not the full spec/status with all the vulnerability data.

This drops memory usage significantly at startup with existing VulnerabilityReports:

| Implementation | Memory (1000 VulnerabilityReports) |
|----------------|-------------------------------------|
| No optimization | ~1002Mi |
| Cache transform | ~825Mi |
| Metadata watch (this PR) | ~317Mi |

The controller now uses `PartialObjectMetadata` for Get/List operations and configures the cache with `UnsafeDisableDeepCopy` since we treat the objects as read-only.

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -26,10 +26,12 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/nats-io/nats.go"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


pohanhuangtw (approved):
LGTM, great optimization technique!

flavio (approved):
Nice!

PR #791, opened 2026-01-24
feat(nats): add StorageClass configuration support

### LINKED ISSUE #556: Feature Request: allow configuration of StorageClass used by NATS
### Is your feature request related to a problem?

Currently there's no way to specify which StorageClass is going to be used by the NATS cluster installed by our helm chart.

As a result of that, the default storage class is going to be used.

### Solution you'd like

It would be great to change our helm chart to have an optional value that can be used to specify which StorageClass is used by the NATS deployment.

### Alternatives you've considered

_No response_

### Anything else?

_No response_

ignorant05:
## Description

Add ability to specify storage class for NATS JetStream file storage

Fix #556 

## Test

### Note: 
The modifications in some indentations are due to my `helm_ls`, i didn't notice them, not before making the PR.

### DIFF
--- a/charts/sbomscanner/values.yaml
+++ b/charts/sbomscanner/values.yaml
@@ -108,6 +108,12 @@ nats:
         secretName: sbomscanner-nats-routes-tls
     jetstream:
       enabled: true
+      fileStorage:
+        enabled: true
+        size: 10Gi
+        storageClassName: ""
+        accessModes:
+          - ReadWriteOnce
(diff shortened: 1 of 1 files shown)


flavio:
@alegrey91 I think it's because of the issue I created (it's linked in the PR). However, I think my issue might be considered invalid.

@fabriziosestito you had some opinion about the issue, but I forgot about that :sweat:

PR #805, opened 2026-02-02
docs(rfc): add WorkloadScan RFC

### LINKED ISSUE #746: docs: write an RFC about the Workload Scan feature

fabriziosestito:
## Description

adds workload scan RFC

Fix #746

### DIFF
--- a/docs/rfc/0007_workload_scan.md
+++ b/docs/rfc/0007_workload_scan.md
@@ -1,0 +1,372 @@
+| Field        | Value                                                      |
+| :----------- | :--------------------------------------------------------- |
+| Feature Name | Workload Scan                                              |
+| Start Date   | 2025-02-01                                                 |
+| Category     | Architecture                                               |
+| RFC PR       | [#XXX](https://github.com/kubewarden/sbomscanner/pull/XXX) |
+| State        | **DRAFT**                                                  |
+
+# Summary
+
+[summary]: #summary
+
+This RFC introduces the WorkloadScan feature, which automatically discovers container images from running workloads and creates the necessary scanning infrastructure. Instead of requiring users to manually configure registries and image filters for runtime scanning, the system watches running pods and generates `Registry` resources on the fly. It also produces `WorkloadScanReport` resources that aggregate vulnerability findings per workload.
+
+# Motivation
+
+[motivation]: #motivation
+
+The existing `Registry` configuration approach works well for users who want to scan entire registries or specific repositories. However, some users want a different perspective: they want to know which vulnerabilities affect their actual running pods, not everything that exists in a registry.
+
+The WorkloadScan feature addresses this use case by:
+
+- Automatically discovering container images from running workloads
+- Creating and managing `Registry` resources scoped to what's actually deployed
+- Producing per-workload vulnerability reports that answer "what needs to be fixed in this deployment?"
+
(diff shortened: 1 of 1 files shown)


docs/rfc/0007_workload_scan.md
```diff
@@ -0,0 +1,340 @@
...
+- A pod is created, updated, or deleted (filtered to only trigger when container images change)
+- A managed `Registry` resource changes (to reconcile it back to the expected state)
+- A managed `WorkloadScanReport` resource changes (to reconcile it back to the expected state)
+- The `WorkloadScanConfiguration` resource changes
+- A namespace's labels change (which may affect selector matching)
+
+The reconciliation logic proceeds as follows:
+
+1. Check if the `WorkloadScanConfiguration` exists. If not, scanning is disabled.
+2. If a `namespaceSelector` is configured, verify the namespace matches. Skip if it doesn't.
```
flavio: We also have to remove workloadscans from repos that no longer match with the selector criteria

docs/rfc/0007_workload_scan.md
```diff
@@ -0,0 +1,340 @@
...
+
+The `ImageAnnotationReconciler` maintains bidirectional references between `Image` resources and the workloads that use them. This enables efficient querying of "which workloads use this image?" without scanning all `WorkloadScanReport` resources.
+
+The reconciler watches both `Image` and `WorkloadScanReport` resources that have the `sbomscanner.kubewarden.io/workloadscan: "true"` label. When an `Image` changes, it finds all `WorkloadScanReport` resources that reference it and updates annotations on the `Image` with workload metadata.
+
+Annotations use the format:
+
+```yaml
+annotations:
+  sbomscanner.kubewarden.io/workloadscan-<uid>: '{"name":"nginx","namespace":"production","containers":2,"summary":{...}}'
```
flavio: what about using `workloadscan.sbomscanner.kubewarden.io/<uid>` as label? The UID already has `-` symbols. I know it's easy to split it but this could look a bit cleaner, doesn't it?

docs/rfc/0007_workload_scan.md
fabriziosestito: Also, worklodscanreports are owned by the respective workload, I Forgot to add this

williamshen9999:
Hi @fabriziosestito , I got two questions. (I'm not sure if it's suitable to have these answers in RFC; let me know if it's not suitable :) )

Here's the first one.

**Question1**:
If the vulnerabilityReport for a specific image is broken (e.g., the format is corrupted or the report is empty), how will this affect the generation of the WorkloadScanReport?
Specifically, will there be any divergent behaviors or error-handling patterns compared to a normal scanning process? (I've gone through the RFC but I don't find the answer. )


Here's the second one.

Scenario:

There is a running Pod based on an image whose vulnerabilityReport already identifies a specific vulnerability (CVE-1234).

A file containing a different vulnerability (CVE-5678) is manually copied into this running Pod (e.g., via kubectl cp).

Subsequently, WorkloadScanConfiguration is enabled for this workload.

**Question2**:
Will the resulting WorkloadScanReport include CVE-5678? ( I've gone through the RFC and I think the answer is "No".  Am I right?)

flavio (approved):

PR #812, opened 2026-02-04
feat(controller): add workloadscan controller

fabriziosestito:
## Description

This PR adds the Workload Scan controller (RFC #805) and introduces setup and helper functions to optimize caching.

References: 
- https://github.com/kubewarden/sbomscanner/blob/feat/workloadscan/docs/rfc/0007_workload_scan.md#reconcilers
- https://github.com/kubewarden/sbomscanner/blob/feat/workloadscan/docs/rfc/0007_workload_scan.md#controller-cache-optimization

It also adds:
- checks to prevent the user to modify manged reosurces 
- enabled field to the WorkloadScanConfiguration

### DIFF
--- a/internal/controller/workloadscan_controller_registry.go
+++ b/internal/controller/workloadscan_controller_registry.go
@@ -107,14 +107,14 @@ func (r *WorkloadScanReconciler) updateRegistry(
 ) (bool, error) {
 	logger := log.FromContext(ctx)
 
-	neededConditions := buildNeededConditions(repositories)
-	oldConditions := extractConditionsFromRegistry(registry)
+	neededMatchConditions := buildNeededMatchConditions(repositories)
+	oldMatchConditions := extractMatchConditionsFromRegistry(registry)
 
-	processedRepositories, processedConditions := r.processExistingRepositories(registry, sourceNamespace, neededConditions)
+	processedRepositories, processedMatchConditions := r.processExistingRepositories(registry, sourceNamespace, neededMatchConditions)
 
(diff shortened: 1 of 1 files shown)


internal/apiserver/admission/workloadscanreport.go
```diff
@@ -0,0 +1,108 @@
...
+	switch attrs.GetOperation() {
+	case admission.Create:
+		obj := attrs.GetObject()
+		if obj == nil {
+			return nil
+		}
+		var ok bool
+		report, ok = obj.(*storagev1alpha1.WorkloadScanReport)
+		if !ok {
+			return nil
```
flavio: nit: should we log that somebody sent our webhook an "unknown" object?

internal/controller/workloadscan_controller.go
```diff
@@ -0,0 +1,191 @@
...
+		},
+	); err != nil {
+		return fmt.Errorf("failed to list managed registries: %w", err)
+	}
+
+	for i := range registries.Items {
+		registry := &registries.Items[i]
+		logger.Info("Deleting managed registry", "registry", registry.Name, "namespace", registry.Namespace)
+		if err := r.Delete(ctx, registry); err != nil && !apierrors.IsNotFound(err) {
+			return fmt.Errorf("failed to delete registry %s/%s: %w", registry.Namespace, registry.Name, err)
```
flavio: Same here, if something goes wrong are we ever going back to remove the leftovers?

PR #846, opened 2026-02-17
fix(helm): nats securitycontext

### LINKED ISSUE #819: SBOMscanner installation fails in some cases due to nats wrong permissions
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Troubleshooting logs

- [x] I reviewed the troubleshooting logs and confirmed they contain no sensitive information.

### Current Behavior

In some cases, we encountered this error, where the SBOMscanner installation remains pending.
The error we get is:
```
Can't start JetStream: could not create storage directory - mkdir /data/jetstream: permission denied
```
This error happened in very limited cases, especially with self-hosted platforms, but we still don't know what the differenciator is.
The cause might be related to the override of the security context applied through the NATS helm chart.
The workaround for now is to remove our security context and leave the ones inherited from the original NATS helm chart.
Here's the command to install it without overriding the NATS security context:
```
helm install sbomscanner kubewarden/sbomscanner   \
  --namespace sbomscanner   \
  --create-namespace   \
  --set controller.replicas=1   \
  --set storage.replicas=1   \
  --set storage.postgres.cnpg.instances=1   \
  --set worker.replicas=1 \
  --set nats.container.merge=null \
  --set nats.reloader.merge=null \
  --wait
```

### Expected Behavior

_No response_

### Steps To Reproduce

I've encountered this error in our internal platform.

### Environment

```markdown
- OS:
- Architecture:
```

### Anything else?

_No response_

alegrey91:
## Description

SBOMscanner's default `nats.container.merge` and `nats.reloader.merge` overrides replace the NATS security context entirely, stripping the `fsGroup` from the pod spec.
Without `fsGroup: 1001`, Kubernetes does not chown the mounted volume to the NATS user (UID 1001) at mount time, so NATS cannot write to `/data/jetstream`.
Instead of nulling the merge (which removes all security hardening), we explicitly set the correct security context values that match what NATS expects (UID/GID 1001), and add `fsGroup: 1001` at the pod level via `podTemplate.merge`.
Fix #819 

## Test

### DIFF
--- a/charts/sbomscanner/values.schema.json
+++ b/charts/sbomscanner/values.schema.json
@@ -154,20 +154,6 @@
                                         },
                                         "readOnlyRootFilesystem": {
                                             "type": "boolean"
-                                        },
-                                        "runAsNonRoot": {
-                                            "type": "boolean"
-                                        },
-                                        "runAsUser": {
-                                            "type": "integer"
-                                        },
-                                        "seccompProfile": {
-                                            "type": "object",
-                                            "properties": {
-                                                "type": {
-                                                    "type": "string"
-                                                }
-                                            }
                                         }
                                     }
                                 }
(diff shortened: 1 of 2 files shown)


flavio (approved):
I think this looks good and sane to me.

Maybe, one way to reduce a bit the repetition could be to try something like that:

```yaml
podTemplate:
  merge:
    spec:
      securityContext:
        runAsUser: 1001
        runAsGroup: 1001
        runAsNonRoot: true
        fsGroup: 1001
        fsGroupChangePolicy: OnRootMismatch
        seccompProfile:
          type: RuntimeDefault

container:
  merge:
    securityContext:
      allowPrivilegeEscalation: false
      readOnlyRootFilesystem: true
      capabilities:
        drop:
        - "ALL"

reloader:
  merge:
    securityContext:
      allowPrivilegeEscalation: false
      readOnlyRootFilesystem: true
      capabilities:
        drop:
        - "ALL"
```

I think this should work, but I cannot test it right now.

flavio (approved):

PR #892, opened 2026-03-06
fix(storage): generate OpenAPI model names

fabriziosestito:
## Description

Add `+k8s:openapi-model-package` tag and `--output-model-name-file` to codegen, so OpenAPI model names use dot-separated API group paths instead of Go import paths. 

This avoids JSON-pointer escaping of `/` in `$ref` strings, which is broken in `kube-openapi`, see: kubernetes/kube-openapi#576.

Related to rancher/rancher#53988

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -506,6 +506,10 @@ linters:
       - path: 'api/.*\.go'
         linters:
           - dupl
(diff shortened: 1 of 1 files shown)


fabriziosestito:
cc @ericpromislow

PR #905, opened 2026-03-06
fix(storage): allow deletion of WorkloadScanReports from GC serviceaccount

fabriziosestito:
## Description

Allow deletion of WorkloadScanReports from the k8s GC serviceaccount.


alegrey91 (approved):
LGTM!

PR #922, opened 2026-03-10
docs: node scan rfc

### LINKED ISSUE #889: RFC: Define how Node Scanning works
Provide an exhaustive description of how the Node Scanning feature will work and how it will integrate with SBOMscanner.

alegrey91:
## Description

Node Scan feature RFC.
Fix #889 

## Test


flavio (changes requested):
This is a good start. There are details that are not covered, such as:

- The daemonset will stay idle most of the time
- How is the scan initiated
- How is the SBOM performed. We both know this is going to be based on trivy scanning the filesystem of the host, mounted into the container. Explain that. Also, we will probably need to provide a way to exclude some host directories from the scan. AFAIK there's the risk of trivy scanning the directories where the container root filesystem is splatted, leading to assigning CVEs of the container images to the host itself.
- Alternative approach: we could find creative way to spin up and down the daemonsets on demand. But this something that requires more work compared to the initial proposal. We expect the agent to be idle most of the time, consuming very little resources. We can revisit the approach later on, when we have real data.

docs/rfc/0008_node_scan.md
```diff
@@ -0,0 +1,165 @@
...
+
+As for the `WorkloadScan` status conditions, the mechanism works the same.
+When `Scheduled` is `true`, then all the other conditions are `false` and their reason is `Scheduled`. 
+When `Pending` is `true`, then all the other conditions are `false` and their reason is `Pending`.
+When `InProgress` is `true`, then all the other conditions are `false` and their reason is `InProgress`.
+When `Complete` is `true`, then all the other conditions are also `false` and their reason is `Complete`.
+
+## Garbage Collection
+
+Garbage collection is crucial to prevent resource orphaning and to maintain a clean cluster state.
```
fabriziosestito, line 149: There are two distinct cleanup mechanisms to consider: the GC and the configuration cleanup.

**Kubernetes garbage collection (Node deletion)**

The owner reference chain is: Node → NodeScanJob, and Node → NodeSBOM → NodeVulnerabilityReport. When a Node is deleted, Kubernetes garbage collection cascades through the owner references and cleans up all node-related resources automatically.

**Reconciler cleanup (NodeScanConfiguration disabled/deleted)**

When the NodeScanConfiguration is disabled or removed, the reconciler must actively clean up NodeScanJobs and NodeSBOMs. NodeVulnerabilityReports are cascade-deleted for free since they are owned by their respective NodeSBOM.

docs/rfc/0008_node_scan.md
```diff
@@ -0,0 +1,165 @@
...
+For this feature we are going to add the following CRDs:
+
+* `NodeScanConfiguration`: Defines the global scan settings.
+  * `scanInterval`: Duration between automated scans.
+    If not specified, the `NodeScanJob` doesn't start.
+  * `nodeSelector`: Filter which nodes are scanned.
+    If not specified, all the nodes are scanned.
+  * `skip`: A list of file/directory paths to be ignored.
+
+* `NodeScanJob`: Represents a single execution of a node scan.
```
fabriziosestito, line 74: Let's reuse the same retention mechanism used for ScanJobs: keep only the latest 10 NodeScanJobs per node.
See: https://github.com/kubewarden/sbomscanner/blob/e5d13da6b7655bf64aafd832c2579ab5b1da1551/docs/rfc/0002_scan_trigger.md?plain=1#L314

docs/rfc/0008_node_scan.md
```diff
@@ -0,0 +1,165 @@
...
+`NodeSBOM` and `NodeVulnerabilityReport` are equal to the [`SBOM`](https://github.com/kubewarden/sbomscanner/blob/main/api/storage/v1alpha1/sbom_types.go) and 
+[`VulnerabilityReport`](https://github.com/kubewarden/sbomscanner/blob/main/api/storage/v1alpha1/vulnerabilityreport_types.go) resource, execept for except for [`ImageMetadata`](https://github.com/kubewarden/sbomscanner/blob/main/api/storage/v1alpha1/image_metadata.go).
+In this case, we are going to use the `NodeMetadata` structure to store 
+information about the node.
+
+`NodeMetadata` will have the following attributes:
+
+* `Name` specifies the unique name of the node in the cluster.
+* `Platform` specifies the OS + CPU architecture of the node. Example: linux/amd64, linux/arm64.
+
```
fabriziosestito, line 96: Please add a section describing which reconcilers need to be introduced, along with a couple of paragraphs explaining their role in the design.

* `NodeScanJobReconciler`
* `NodeScanReconciler`
* `NodeScanRunner` (as a [runnable](https://github.com/kubewarden/sbomscanner/blob/e5d13da6b7655bf64aafd832c2579ab5b1da1551/docs/rfc/0002_scan_trigger.md?plain=1#L321))

The `NodeScanRunner` filters nodes using the `nodeSelector` and platform criteria. The platform filter can be implemented using controller-runtime indexes.

docs/rfc/0008_node_scan.md
```diff
@@ -0,0 +1,165 @@
...
+To trigger a new scan, the user can set the `scanInterval` on the `NodeScanConfiguration`,
+or leave the `scanInterval` not set and apply a `NodeScanJob` manually 
+(with the `NodeScanConfiguration` already applied) as we already do for the `Registry`.
+
+Please, note that `NodeScanConfiguration` is a singleton resource, 
+meaning that there can be only one instance of it in the cluster.
+
+## CRDs
+
+For this feature we are going to add the following CRDs:
```
fabriziosestito, line 65: please add example CRDs

docs/rfc/0008_node_scan.md
```diff
@@ -0,0 +1,165 @@
...
+
+* `NodeScanConfiguration`: Defines the global scan settings.
+  * `scanInterval`: Duration between automated scans.
+    If not specified, the `NodeScanJob` doesn't start.
+  * `nodeSelector`: Filter which nodes are scanned.
```
fabriziosestito, line 70: add platform(s) filter

docs/rfc/0008_node_scan.md
```diff
@@ -0,0 +1,97 @@
...
+The worker will be provided with a flag to operate between image scanning and node scanning 
+modes, allowing for significant code reuse across different scan targets.
+
+## CRDs
+
+For this feature we are going to add the following CRDs:
+
+* `NodeScanConfiguration`: Defines the global scan settings.
+  * `scanInterval`: Duration between automated scans.
+  * `skip`: A list of file/directory paths to be ignored.
```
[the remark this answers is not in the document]
fabriziosestito: Yes, I would go with the singleton CRD, please let's specify this in the RFC

docs/rfc/0008_node_scan.md
```diff
@@ -0,0 +1,165 @@
...
+## CRDs
+
+For this feature we are going to add the following CRDs:
+
+* `NodeScanConfiguration`: Defines the global scan settings.
+  * `scanInterval`: Duration between automated scans.
+    If not specified, the `NodeScanJob` doesn't start.
+  * `nodeSelector`: Filter which nodes are scanned.
+    If not specified, all the nodes are scanned.
+  * `skip`: A list of file/directory paths to be ignored.
```
fabriziosestito, line 72: We could use something like this for skip patterns:

```yaml
 # Gitignore-style patterns to exclude from filesystem scans.
  # Trailing "/" = directory, otherwise = file.
  skipPatterns:
    - "node_modules/"       # → --skip-dirs node_modules
    - "**/vendor/"          # → --skip-dirs (glob expanded at scan time)
    - ".git/"               # → --skip-dirs .git
    - "*.min.js"            # → --skip-files *.min.js
    - "package-lock.json"   # → --skip-files package-lock.json
```

docs/rfc/0008_node_scan.md
```diff
@@ -0,0 +1,165 @@
...
+information about the node.
+
+`NodeMetadata` will have the following attributes:
+
+* `Name` specifies the unique name of the node in the cluster.
+* `Platform` specifies the OS + CPU architecture of the node. Example: linux/amd64, linux/arm64.
+
+## Scan Workflow
+
+1. The user applies a `NodeScanConfiguration` with a defined `scanInterval` or applies a `NodeScanJob` manually.
```
fabriziosestito, line 99: Let's revisit this and add the runner. That way we have two branches in the flow: one triggered by the user, and one triggered by the runner.

fabriziosestito (changes requested):

docs/rfc/0008_node_scan.md
alegrey91: done

docs/rfc/0008_node_scan.md
alegrey91: done

docs/rfc/0008_node_scan.md
```diff
@@ -0,0 +1,215 @@
...
+apiVersion: sbomscanner.kubewarden.io/v1alpha1
+kind: NodeScanJob
+metadata:
+  name: node-scan-1
+```
+
+* `NodeSBOM`: Stores the Software Bill of Materials for a specific node.
+
+* `NodeVulnerabilityReport`: Contains the results of the vulnerability analysis.
+
```
fabriziosestito, line 125: It is worth pointing out that these two resources are stored in the Storage API Extension Server and are very similar to their counterparts, `SBOM` and `VulnerabilityReport`.

docs/rfc/0008_node_scan.md
fabriziosestito: @alegrey91 this has not been addressed yet. The RFC does not clearly distinguish between GC and reconciler cleanup. Please add the clarification above.

docs/rfc/0008_node_scan.md
```diff
@@ -0,0 +1,215 @@
...
+## Reconcilers
+
+The NodeScan reconciler is responsible for creating `NodeScanJob` resources based on the 
+`NodeScanConfiguration` and ensuring that the scan jobs are executed according to the defined schedule.
+
+The NodeScanJob reconciler is responsible for managing the lifecycle of `NodeScanJob` resources, 
+including checking for concurrent scans, updating status conditions, and triggering the scan on 
+the node by publishing a message to the NATS subject `sbomscanner.nodescan.{node-name}`.
+
+the NodeScan runner (as a runnable) is responsible for executing the scan on the node automatically,
```
fabriziosestito:
```suggestion
The NodeScan runner (as a runnable) is responsible for executing the scan on the node automatically, 
```

fabriziosestito: also please link to the runnable k8s documentation

docs/rfc/0008_node_scan.md
fabriziosestito: @alegrey91 there is currently no mention anywhere of the `.gitignore`-style skip pattern implementation. Either keep the YAML comments above, or add a short paragraph explaining how we intend to implement these patterns and how this works in Trivy.

fabriziosestito (changes requested):

docs/rfc/0008_node_scan.md
alegrey91: done, thanks for pointing it out.

fabriziosestito (approved):

docs/rfc/0008_node_scan.md
```diff
@@ -0,0 +1,242 @@
...
+For this feature we are going to add the following CRDs:
+
+* `NodeScanConfiguration`: Defines the global scan settings.
+  * `scanInterval`: Duration between automated scans.
+    If not specified, the `NodeScanJob` doesn't start.
+  * `nodeSelector`: Filter which nodes are scanned.
+    If not specified, all the nodes are scanned.
+  * `skipPatterns`: A list of file/directory paths to be ignored.
+    This can be expressed as .gitignore-like patterns (eg. `**/tmp/**` to ignore all the `tmp` directories).
+    If not specified, no file is ignored.
```
flavio, line 74: Are we planning to have a list of directories that is always ignored, regardless of the user input?

I **think** trivy can get confused by looking at the directory where the root filesystems of containers are unpacked. Also, what about `/dev`, `/proc`, `/sys`,...

flavio (approved):
LGTM, I left just a minor question

PR #1026, opened 2026-04-10
feat(helm): Configure service type for mcp

hierynomus:
## Description

This PR adds support for configuring the service type for the MCP service, allowing it to be exposed through a NodePort.

## Test
n/a

### Tradeoff

None
## Checklist


fabriziosestito:
Closinng as not planned for now, we will think of a better integration with Rancher in the following milestones.

PR #1075, opened 2026-04-27
feat(helm): expose CNPG Cluster imageName

### LINKED ISSUE #1055: Feature Request: Expose imageName field for CNPG Cluster resource in Helm values
### Is your feature request related to a problem?

Yes. In restricted or air-gapped environments, all container images must be pulled from a local private registry. While the Helm chart allows specifying a private registry for the CNPG operator deployment, it currently lacks a mechanism to propagate a custom image path to the Cluster resources created by the operator.

As a result, while the operator starts correctly, the resulting PostgreSQL pods fail to pull because they default to a public or hardcoded registry that is unreachable in these environments.

### Solution you'd like

I would like the Helm chart to expose the imageName field for the Cluster resource definition. This would allow users to define a custom repository and tag for the database instances directly in values.yaml

### Alternatives you've considered

I have tried configuring the operator's image repository, but as per the CNPG architecture, the operator's image and the managed instances' images are managed separately. Manually editing the Cluster CRD after deployment is not sustainable for GitOps workflows.

### Anything else?

This is a common requirement for enterprise users and and air-gapped scenarios where public registry access (like Docker Hub or GHCR) is disabled. Providing this toggle in the Helm chart ensures the deployment is fully portable across different registry configurations.

fabriziosestito:
## Description

Fix #1055

### DIFF
--- a/charts/sbomscanner/values.yaml
+++ b/charts/sbomscanner/values.yaml
@@ -57,6 +57,11 @@ storage:
     cnpg:
       enabled: true
       instances: 3
+      # Override the PostgreSQL container image used by the CNPG Cluster.
+      # Useful for air-gapped environments where the default image is not reachable.
+      # Must be a full image reference (e.g. `my-registry.example.com/postgresql:18.1`).
+      # If empty, CNPG uses its default operand image.
+      imageName: ""
       storage:
         # Size of the storage. Required if not already specified in the PVC template.
         # Changes to this field are automatically reapplied to the created PVCs.
--- a/docs/installation/helm-values.md
+++ b/docs/installation/helm-values.md
@@ -87,6 +87,7 @@ storage:
     cnpg:
       enabled: true
       instances: 3
+      imageName: ""
       storage:
         size: 1Gi
         resizeInUseVolumes: true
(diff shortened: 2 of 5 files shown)


charts/sbomscanner/values.yaml
```diff
@@ -57,6 +57,11 @@ storage:
     cnpg:
       enabled: true
       instances: 3
+      # Override the PostgreSQL container image used by the CNPG Cluster.
+      # Useful for air-gapped environments where the default image is not reachable.
+      # Must be a full image reference (e.g. `my-registry.example.com/postgresql:18.1`).
+      # If empty, CNPG uses its default operand image.
```
flavio, line 63:
```suggestion
      # If empty, CNPG uses its default operator image.
```

docs/installation/helm-values.md
```diff
@@ -96,6 +97,7 @@ storage:
 
 **Configuration options:**
 - `instances`: Number of PostgreSQL replicas (default: 3)
+- `imageName`: Full image reference for the PostgreSQL container (e.g. `my-registry.example.com/postgresql:18.1`). Useful for air-gapped environments where the default CNPG operand image is not reachable. If empty, CNPG uses its default operand image.
```
flavio, line 100:
```suggestion
- `imageName`: Full image reference for the PostgreSQL container (e.g. `my-registry.example.com/postgresql:18.1`). Useful for air-gapped environments where the default CNPG operand image is not reachable. If empty, CNPG uses its default operator image.
```

flavio (changes requested):
LGTM, just a minor fix

docs/installation/helm-values.md
fabriziosestito: `imageName`specifies the operand image, not the operator image. See: https://cloudnative-pg.io/documentation/1.20/operator_capability_levels/#override-of-operand-images-through-the-crd

charts/sbomscanner/values.yaml
fabriziosestito: `imageName` specifies the operand image, not the operator image. See: https://cloudnative-pg.io/documentation/1.20/operator_capability_levels/#override-of-operand-images-through-the-crd

flavio (approved):

alegrey91 (approved):

Issue #1071, opened 2026-04-27
SBOMscanner v0.11.0 release


fabriziosestito:
### Checklist

This issue tracks the release of a new version of SBOMscanner. Please
follow the checklist below to ensure a smooth release process.

- [x] Notify consumers if any.
- [x] Tag desired version in [kubewarden/sbomscanner](https://github.com/kubewarden/sbomscanner) repo.
  `git tag -s -a -m "v0.11.0" v0.11.0`
- [ ] Push the tag upstream.
  `git push upstream v0.11.0`
- [x] Wait for CI to succeed.
- [ ] Review the PR opened by Updatecli that bump the Helm Chart, then **Squash and merge**.
- [x] Trigger an [update-sbomscanner](https://github.com/kubewarden/helm-charts/actions/workflows/update-sbomscanner.yaml)
  job run. This opens an automated PR that syncs sbomscanner charts in [kubewarden/helm-charts](https://github.com/kubewarden/helm-charts) repo.
- [x] Merge automated PR on kubewarden/helm-charts repo.
- [x] chart-releaser releases the charts on Helm chart repo.
- [x] Write and release the blog post about the release.
  - [x] Publish posts in Mastodon, Bluesky.

PR #1079, opened 2026-04-28
ci: harden run blocks against template injection

fabriziosestito:
## Description
Hardens run blocks against template injection by using env vars


fabriziosestito:
Blocked by #1071

flavio (approved):

PR #1094, opened 2026-05-04
fix(storage): wait for NATS subscription before publishing in watcher

fabriziosestito:
## Description

`natsWatcher` and `WorkloadScanReportWatcher` expose a `Setup(ctx)` method that runs `nc.Subscribe` followed by `nc.Flush()` synchronously, so the NATS server has acknowledged the `SUB` before `Setup` returns. `Start(ctx)` is a blocking run loop that waits on `ctx.Done()` and unsubscribes on shutdown.

Callers can rely on subscription readiness once `Setup` returns nil, and any subscription error is surfaced immediately instead of being silently dropped.
- Introduce a `Watcher` interface (`manager.Runnable` + `Setup`) and switch the storage registry stores and `StorageAPIServer` to it; `Setup` is invoked before `Start` in the existing errgroup.
- Update store and workloadscanreport watcher tests to call `Setup(ctx)` before `go Start(ctx)` instead of relying on goroutine scheduling + publisher-side `nc.Flush()`.
- Fixes flaky `TestStoreTestSuite/TestHandleMessageNotFoundStillBroadcasts`, `TestWatchResourceVersionZero`, and `TestWatchSpecificResourceVersion`.

### DIFF
--- a/internal/apiserver/storage.go
+++ b/internal/apiserver/storage.go
@@ -21,7 +21,6 @@ import (
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	basecompatibility "k8s.io/component-base/compatibility"
 	baseversion "k8s.io/component-base/version"
-	"sigs.k8s.io/controller-runtime/pkg/manager"
 
 	"github.com/kubewarden/sbomscanner/api/storage/install"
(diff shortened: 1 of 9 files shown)


flavio (dismissed):
LGTM, good catch

flavio (approved):

PR #1135, opened 2026-05-18
feat: add SeveritySource to VulnerabilityReport

fabriziosestito:
## Description
Adds a new `SeveritySource` field on `Vulnerability` that identifies the vendor that produced the severity (`"nvd"`, `"ghsa"`, `"redhat"`, `"alpine"`, …). 
   
The field is populated directly from the upstream scanner's per-finding source, currently passed through from Trivy.

Consumers should use this key to look up the matching entry in the `cvss` map and display that score alongside `Severity`. This mirrors the lookup pattern Trivy uses in its own SARIF reporter.
If there is no match, the consumer can fallback to default CVSS values, see: 

  https://github.com/aquasecurity/trivy/blob/v0.70.0/pkg/report/sarif.go#L447-L460

  ```go
  func severityToScore(severity string) string {
      switch severity {
      case "CRITICAL":
          return "9.5"
      case "HIGH":
          return "8.0"
      case "MEDIUM":
          return "5.5"
      case "LOW":
          return "2.0"
      default:
          return "0.0"
      }
  }
  ```

### DIFF
--- a/api/storage/v1alpha1/vulnerabilityreport_types.go
+++ b/api/storage/v1alpha1/vulnerabilityreport_types.go
@@ -130,6 +130,12 @@ type Vulnerability struct {
 	// Severity rating (e.g., "HIGH", "MEDIUM")
 	Severity string `json:"severity" protobuf:"bytes,10,req,name=severity"`
 
+	// SeveritySource identifies the vendor that produced the Severity
+	// (e.g. "nvd", "ghsa", "redhat", "alpine").
+	// Consumers can use this key to look up the matching entry in the CVSS map to display alongside Severity.
+	// May be empty when the source vendor is not known.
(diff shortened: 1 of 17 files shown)


fabriziosestito:
Blocked, waiting for feedback from @davideiori1

davideiori1:
@fabriziosestito I agree with the approach. Vendor's severity and cvss score are always to be preferred over the  values that are not coming from the vendor itself. However, the same principle applies also when the vendor does not publish a cvss score but only the severity is available: it is better to show a `N/A` score than a made-up value, and stick to the severity provided by the vendor.

alegrey91 (approved):

flavio (approved):

PR #1184, opened 2026-05-28
feat(controller): reject ScanJob creation when referenced Registry does not exist

### LINKED ISSUE #1180: Reject `ScanJob` creation in the webhook when the referenced `Registry` does not exist.
Reject `ScanJob` creation in the webhook when the referenced `Registry` does not exist.

Currently, this is handled by failing in the controller, which should remain in place to avoid TOCTOU issues, but rejecting it upfront in the webhook provides a better user experience.

fabriziosestito:
## Description

Fixes #1180

### DIFF
--- a/internal/webhook/v1alpha1/scanjob_webhook.go
+++ b/internal/webhook/v1alpha1/scanjob_webhook.go
@@ -86,6 +86,20 @@ func (v *ScanJobCustomValidator) ValidateCreate(ctx context.Context, scanJob *v1
 		}
 	}
 
+	registry := &v1alpha1.Registry{}
+	if err := v.client.Get(ctx, client.ObjectKey{
+		Name:      scanJob.Spec.Registry,
+		Namespace: scanJob.GetNamespace(),
+	}, registry); err != nil {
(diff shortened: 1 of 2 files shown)


internal/webhook/v1alpha1/scanjob_webhook.go
```diff
@@ -19,7 +19,7 @@ import (
 func SetupScanJobWebhookWithManager(mgr ctrl.Manager) error {
 	err := ctrl.NewWebhookManagedBy(mgr, &v1alpha1.ScanJob{}).
 		WithValidator(&ScanJobCustomValidator{
-			client: mgr.GetClient(),
+			client: mgr.GetAPIReader(),
```
alegrey91, line 22: is the API reader used intentionally to avoid reading from the cache, right?

fabriziosestito: Yes, this was a mistake in the initial implementation. We need to use the API reader here to avoid stale cache reads.

alegrey91 (approved):

flavio (approved):

PR #1190, opened 2026-06-03
feat(debug): auto-discover release, collect cnpg + extra CRs

fabriziosestito:
## Description
Auto-discovers the SBOMscanner Helm release and extends `collect` with CNPG logs, init/previous-container logs, an opt-in `--manifests` dump, and an `--output-dir` flag. 

Docs updated accordingly.


flavio (approved):
nice improvements!

alegrey91 (approved):

PR #1214, opened 2026-06-08
feat(telemetry): OTel plumbing

### LINKED ISSUE #1215: OTel: plumbing (internal/telemetry foundation)
Part of #972.

Foundation for OTel work: `internal/telemetry` package wiring `Setup`, slog `trace_id`/`span_id` handler, NATS header carrier, and `Tracer`/`Meter` helpers. Called from each `cmd/*/main.go`. No-op when `OTEL_EXPORTER_OTLP_ENDPOINT` is unset.

fabriziosestito:
## Description

Introduces the shared OpenTelemetry bootstrap for the four sbomscanner binaries, gated on `OTEL_EXPORTER_OTLP_ENDPOINT`. 

When the variable is unset, the binaries install no-op providers and behave exactly as before; when it is set, traces and metrics are exported over OTLP/gRPC to any collector the operator provides.
All runtime configuration is done via the standard [OpenTelemetry environment variables](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/); no new CLI flags are introduced.

The new `internal/telemetry` package holds the `Setup` entry point, an `slog.Handler` wrapper that decorates log records with the active `trace_id` and `span_id` for log-to-trace correlation, a `TextMapCarrier` over NATS headers used by the messaging layer, and small `Tracer` / `Meter` helpers that expose the spec-conformant instrumentation scope. 

Each `cmd/*/main.go` now calls `telemetry.Setup` right after slog init and defers a bounded shutdown. 

Also, it adds build-time version injecitonm, which is used as metadata in the telemetry.

Fixes #1215

### DIFF
--- a/Dockerfile.controller
+++ b/Dockerfile.controller
@@ -15,7 +15,11 @@ COPY api/ api/
 COPY internal/ internal/
 COPY pkg/ pkg/
 
-RUN CGO_ENABLED=0 GOOS=linux GOEXPERIMENT=jsonv2 go build -a -o ./controller ./cmd/controller
+ARG VERSION=dev
+
+RUN CGO_ENABLED=0 GOOS=linux GOEXPERIMENT=jsonv2 go build -a \
+    -ldflags "-X github.com/kubewarden/sbomscanner/internal/version.Version=${VERSION}" \
+    -o ./controller ./cmd/controller
 
 # Use distroless as minimal base image to package the manager binary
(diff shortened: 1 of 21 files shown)


flavio (approved):
that's pretty neat :clap:

alegrey91 (approved):
LGTM. Should we include some words in the doc to inform the user about the possibility of using OTEL?

PR #1297, opened 2026-07-21
feat(webhook): add warning when skippatterns is not set

### LINKED ISSUE #1294: return a warning when applying a `NodeScanConfiguration` with `skipPatterns` set
When a user define its own `skipPattern` strategy, the default one is overrided completly.
This is expected as the proper behaviour, but we should make the user aware about this and suggest him to add `skipPatterns` that include `procfs` paths (eg. `/run/k3s/containerd/`).

alegrey91:
## Description

This PR adds a warning message from the webhook when a user sets a `spec.skipPatterns` list different from the default one. This is needed because removing the default list can cause errors with the node scan.
Fix #1294 

## Test

## Checklist

- [ ] I have read and understood the [Kubewarden AI Policy](https://github.com/kubewarden/community/blob/main/AI_POLICY.md)


api/v1alpha1/nodescanconfiguration_types_test.go
```diff
@@ -0,0 +1,34 @@
...
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"sigs.k8s.io/yaml"
+)
+
+// TestDefaultSkipPatternsMatchCRD guards against drift between DefaultSkipPatterns
+// and the +kubebuilder:default marker on NodeScanConfigurationSpec.SkipPatterns.
+// It parses the generated CRD and asserts its schema default equals the Go slice,
+// which also fails if the CRD was not regenerated after a marker change.
+// Path is relative to this package; update it if the chart layout moves.
+func TestDefaultSkipPatternsMatchCRD(t *testing.T) {
```
fabriziosestito: This is a smart hack for testing purposes, but I wouldn’t go with this approach.
We should move the validation logic from the CRD Kubebuilder tag to the validating webhook instead.

internal/webhook/v1alpha1/nodescanconfiguration_webhook.go
```diff
@@ -16,6 +17,11 @@ import (
 	"github.com/kubewarden/sbomscanner/api/v1alpha1"
 )
 
+const skipPatternsOverrideWarning = "spec.skipPatterns overrides the default skip patterns. " +
+	"The defaults exclude container-runtime state (e.g. /run/containerd, /run/k3s/containerd) " +
+	"which exposes the procfs of running containers on the node. " +
+	"Scanning those paths may cause the node scan to fail."
+
```
fabriziosestito: Since the check is now about inclusion rather than an exact match, I'd name this after what actually triggers the warning (defaults missing, not merely overridden). I'd also drop the procfs detail and render the full default list from `v1alpha1.DefaultSkipPatterns` instead of hand-picking examples, so the warning can never drift out of sync with the actual defaults:

```suggestion
  func missingDefaultSkipPatternsWarning() string {
        return fmt.Sprintf("spec.skipPatterns does not include all of the default skip patterns (%s). "+
                "The defaults exclude container-runtime state directories; scanning those paths may cause the node scan to fail.
                strings.Join(v1alpha1.DefaultSkipPatterns, ", "))
  }
  ```

File: internal/webhook/v1alpha1/nodescanconfiguration_webhook.go [hunk 3]
```diff
@@ -66,7 +72,30 @@ func (v *NodeScanConfigurationCustomValidator) ValidateUpdate(_ context.Context,
...
+}
+
+// skipPatternsWarnings returns a warning when the user overrides the default
+// skipPatterns, since removing the container-runtime paths can make the node
+// scan fail while walking a running container's procfs.
+func skipPatternsWarnings(config *v1alpha1.NodeScanConfiguration) admission.Warnings {
+	if config.Spec.SkipPatterns == nil {
+		return nil
+	}
+	if slices.Equal(*config.Spec.SkipPatterns, v1alpha1.DefaultSkipPatterns) {
```
fabriziosestito: Rather than requiring an exact match with Equal, we could check for inclusion, so the condition still holds when the user has added their own patterns alongside the defaults:
```suggestion
	if sets.New(*config.Spec.SkipPatterns...).HasAll(v1alpha1.DefaultSkipPattern...) {
```

fabriziosestito (changes requested):

api/v1alpha1/nodescanconfiguration_types_test.go
alegrey91: This was my first thought for this PR, but then I changed my mind.
If the webhooks are down for any reason, we are going to lose the default skip patterns.
I would prefer to keep it on the apiserver as we have now with kubebuilder.

fabriziosestito: I insist on going through the webhook route. If the webhooks are down for any reason, the request will be rejected with a `500` anyway, so that's not a good reason to keep this validation in the CRD. We already perform other validations (such as platform validation) in the webhooks, so this should be handled there as well.

Please update this accordingly.

api/v1alpha1/nodescanconfiguration_types_test.go
```diff
@@ -0,0 +1,34 @@
+package v1alpha1
```
fabriziosestito: this test needs to be removed

internal/webhook/v1alpha1/nodescanconfiguration_webhook.go
```diff
@@ -50,7 +65,7 @@ func (v *NodeScanConfigurationCustomValidator) ValidateCreate(_ context.Context,
 		)
 	}
 
-	return nil, nil
+	return nodeScanConfigurationWarnings(config), nil
```
fabriziosestito: we want to return both warnings and errors. 

```suggestion
func (v *NodeScanConfigurationCustomValidator) ValidateCreate(_ context.Context, config *v1alpha1.NodeScanConfiguration) (admission.Warnings, error) {
	v.logger.Info("Validation for NodeScanConfiguration upon creation", "name", config.GetName())

	warnings, allErrs := validateNodeScanConfiguration(config)

	if len(allErrs) > 0 {
		return warnings, apierrors.NewInvalid(
			v1alpha1.GroupVersion.WithKind("NodeScanConfiguration").GroupKind(),
			config.Name,
			allErrs,
		)
	}

	return warnings, nil
```

internal/webhook/v1alpha1/nodescanconfiguration_webhook.go
```diff
@@ -66,7 +81,31 @@ func (v *NodeScanConfigurationCustomValidator) ValidateUpdate(_ context.Context,
...
+
+// nodeScanConfigurationWarnings aggregates all admission warnings for a
+// NodeScanConfiguration. Add new field warnings here.
+func nodeScanConfigurationWarnings(config *v1alpha1.NodeScanConfiguration) admission.Warnings {
+	var warnings admission.Warnings
+
+	warnings = append(warnings, skipPatternsWarnings(config)...)
+
+	return warnings
+}
```
fabriziosestito: This function is superfluous

internal/webhook/v1alpha1/nodescanconfiguration_webhook.go
```diff
@@ -66,7 +81,31 @@ func (v *NodeScanConfigurationCustomValidator) ValidateUpdate(_ context.Context,
...
+// container's directory tree.
+func skipPatternsWarnings(config *v1alpha1.NodeScanConfiguration) admission.Warnings {
+	if config.Spec.SkipPatterns == nil {
+		return nil
+	}
+	if sets.New(*config.Spec.SkipPatterns...).HasAll(v1alpha1.DefaultSkipPatterns...) {
+		return nil
+	}
+	return admission.Warnings{missingDefaultSkipPatternsWarning()}
 }
```
fabriziosestito:
```suggestion
// validateNodeScanSkipPatterns returns a warning when the user-supplied skipPatterns
// list drops one or more of the default patterns, since removing the
// container-runtime paths can make the node scan fail while walking a running
// container's directory tree.
func validateNodeScanSkipPatterns(skipPatterns []string) admission.Warnings {
	if sets.New(skipPatterns...).HasAll(v1alpha1.DefaultSkipPatterns...) {
		return nil
	}
	return admission.Warnings{missingDefaultSkipPatternsWarning()}
}
```

fabriziosestito: Can you also move this function at the bottom together with the other validation functions

internal/webhook/v1alpha1/nodescanconfiguration_webhook.go
fabriziosestito:
```suggestion
func validateNodeScanConfiguration(config *v1alpha1.NodeScanConfiguration) (admission.Warnings, field.ErrorList) {
	var warnings admission.Warnings
	var allErrs field.ErrorList

	if config.Spec.SkipPatterns != nil {
		warnings = append(warnings, validateNodeScanSkipPatterns(*config.Spec.SkipPatterns)...)
	}

	if err := validateScanInterval(config.Spec.ScanInterval); err != nil {
		allErrs = append(allErrs, err)
	}
	allErrs = append(allErrs, validatePlatforms(config.Spec.Platforms)...)
	allErrs = append(allErrs, validateNodeSelector(config.Spec.NodeSelector)...)

	return warnings, allErrs
}
```

fabriziosestito (changes requested):

alegrey91:
@fabriziosestito I rewrote the commit history since it was not well describing the changes. please review it again when you have time.

fabriziosestito (approved):

fabriziosestito:
Ok to go, let's wait for the runners to run the test suite @alegrey91

fabriziosestito (approved):

PR #1305, opened 2026-07-28
refactor(controller): pass the instrumentation to SetupWithManager instead of a struct field

fabriziosestito:
## Description
Small cleanup to pass the instrumentation instead of using the public field


alegrey91:
LGTM, let's wait for the runners to be back, and then we can merge it

PR #1307, opened 2026-07-29
feat(worker): add traces and metrics

### LINKED ISSUE #1218: OTel: worker traces and metrics
Part of #972.

Handler spans (`CreateCatalog`, `GenerateSBOM`, `ScanSBOM`), NATS consume spans, registry HTTP via `otelhttp`, Trivy invocation spans. Scan duration per stage, vulns by severity, registry/Trivy latency, per-workload and per-image metrics with exemplars.

fabriziosestito:
## Description

This PR instruments the worker, continuing the work started with the controller in #1280. Fixes #1218.

Until now, the job trace stopped at the controller: once a message was published to NATS, we lost track of it. Now the trace ID travels inside the message headers, so when the worker picks up a message it continues the same trace. The result is that a single scan shows up in Jaeger as one tree: the reconcile, then the catalog creation, then the SBOM generation for each image, then the vulnerability scan, with the Trivy runs visible inside each step. When a handler skips a message (for example because the job was deleted in the meantime), the span says why.

The worker also exports metrics now: how long each stage of the pipeline takes, how many images were scanned, how many vulnerabilities were found by severity, plus timings for registry calls and Trivy runs, and a counter for handler errors. The example Grafana dashboard gained a Worker row showing all of this, and the Tilt settings example now enables the telemetry stack by default so it is easy to try locally.

RFC 0009 is updated to describe what actually shipped.

---
image

---

image

---

image

### DIFF
--- a/charts/sbomscanner/templates/controller/deployment.yaml
+++ b/charts/sbomscanner/templates/controller/deployment.yaml
@@ -44,6 +44,14 @@ spec:
             {{- if .Values.observability.otel.endpoint }}
             - name: OTEL_EXPORTER_OTLP_ENDPOINT
               value: {{ .Values.observability.otel.endpoint | quote }}
+            - name: K8S_POD_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.name
+            - name: K8S_POD_NAMESPACE
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
             {{- end }}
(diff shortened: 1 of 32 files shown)


flavio (approved):
I like how the pattern that was used when introducing traces and metrics earlier on fits nicely also with the worker.

The code is pretty elegant.

alegrey91 (approved):

PR #1319, opened 2026-08-04
feat: set lastUpdate and nextUpdate annotations for all the layers

### LINKED ISSUE #1308: Complete sbomscannerdb CLI to satisfy RFC 0010 (enrichment vulnerability database)
Tracks the remaining work to bring the sbomscannerdb CLI in line with the requirements in [RFC 0010](https://github.com/kubewarden/sbomscanner/blob/main/docs/rfc/0010_enrichment_vulnerability_database.md). The CLI builds, pushes, and manages the OCI artifact that distributes auxiliary data (KEV, EPSS, …) to SBOMScanner workers.

Already implemented:
- [x] build: download feeds, one-layer-per-file packaging, tag in local store
- [x] push: push artifact to registry (ORAS)
- [x] pull: pull artifact, extract data layers
- [x] Reproducible tar.gz (normalized mode/mtime/owner, epoch created)
- [x] Empty config blob / OCI 1.1 artifact style
- [x] KEV + EPSS downloaders with validation

## To do
- [x] **`inspect` command**:
RFC lists sbomscannerdb inspect: print the manifest, layers, media types, and annotations. Only a list command (local store refs) exists today, no manifest/annotation inspection.
- [x] **Freshness annotations on the manifest**:
Write on build:
  - `io.kubewarden.sbomscanner.db.lastUpdate` = build time.
  - `io.kubewarden.sbomscanner.db.nextUpdate` = computed from the shortest update cadence among bundled feeds.
These drive the worker's cheap staleness check.
- [x] **Reconcile media-type naming convention**:
RFC proposes format-based, feed-agnostic media types:
  * `application/vnd.sbomscanner.db.file.v1.json+gzip`
  * `application/vnd.sbomscanner.db.file.v1.csv+gzip`
- [x] **Multi-file-per-database support**:
RFC states tar (not bare gzip) is used so a data source made up of multiple files maps to one tar.gz layer. Current packaging assumes one input file per feed, support N-files-into-one-layer.

alegrey91:
## Description

This PR adds the missing requirements needed to complete the sbomscannedb cli.
Specifically, this PR adds the following functionalities:
* inspect: locally and remote with a dedicated command
* reproducible tar.gz (normalized mode/mtime/owner, epoch created)
* support for kev and epss (plus an easily extendible interface)
* manage next update annotations on the manifest (for each layer)
* have a coherent media-type naming convention
* handle multi-file-per-database

Fix #1308 

## Test

## Checklist

- [ ] I have updated the documentation via a pull request in [docs](https://github.com/kubewarden/docs/tree/main/docs/sbom-scanner) repository.


cmd/sbomscannerdb/commands.go
```diff
@@ -5,15 +5,38 @@ import (
...
 	"text/tabwriter"
+	"time"
 
 	"github.com/kubewarden/sbomscanner/internal/sbomscannerdb/datafeed"
 	"github.com/kubewarden/sbomscanner/internal/sbomscannerdb/oci"
 )
 
+// sourceDateEpochEnv is the well-known variable used to pin a build's timestamp
+// for reproducible builds. When set, it overrides the wall-clock build time.
+const sourceDateEpochEnv = "SOURCE_DATE_EPOCH"
```
fabriziosestito: I don't think this logic belongs in the `commands` package. I'd take a different approach:

Have `NewBuilder` accept `sourceDateEpoch` (that might be empty) instead of the window, and let it calculate the window using the `buildTime` helper function. That helper should be moved alongside the builder.

The name itself suggests it belongs there; if it's responsible for "building" something, it should live with the builder.

cmd/sbomscannerdb/buildtime_test.go
```diff
@@ -0,0 +1,49 @@
+package main
```
fabriziosestito: This should be moved next to the builder helper in `oci/artifact.go`. Also, as a general rule, we don't name test files after the functions they test. This test should therefore be moved to `artifact_test.go`.

internal/sbomscannerdb/oci/artifact_test.go
```diff
@@ -0,0 +1,130 @@
+package oci
```
fabriziosestito, line 1: This file should be renamed to `artifact_test.go`

cmd/sbomscannerdb/commands.go
```diff
@@ -119,3 +119,31 @@ func runPull(ctx context.Context, ref string, config oci.Config, logger *slog.Lo
...
+		}
+		view, err = store.Inspect(ctx, ref)
+	} else {
+		view, err = oci.NewRemote(config, logger).Inspect(ctx, ref)
+	}
+	if err != nil {
+		return fmt.Errorf("inspect artifact: %w", err)
+	}
+
+	return render(os.Stdout, view)
```
fabriziosestito: Just render json here without having multiple formats

cmd/sbomscannerdb/inspect_format.go
```diff
@@ -0,0 +1,43 @@
...
+	"encoding/json"
+	"fmt"
+	"io"
+	"sort"
+	"strings"
+
+	"github.com/kubewarden/sbomscanner/internal/sbomscannerdb/oci"
+)
+
+// inspectRenderer writes a manifest view to w in a specific format.
```
fabriziosestito: We agreed on just having json.

cmd/sbomscannerdb/inspect_test.go
```diff
@@ -0,0 +1,63 @@
+package main
```
fabriziosestito: We can remove this test

fabriziosestito (changes requested):

internal/sbomscannerdb/oci/buildtime_test.go
```diff
@@ -0,0 +1,41 @@
+package oci
```
fabriziosestito: this  should go in `artifact_test.go`

internal/sbomscannerdb/oci/artifact_test.go
fabriziosestito, line 1: this should go in `artifact_test.go`

internal/sbomscannerdb/oci/manifest_test.go
```diff
@@ -0,0 +1,22 @@
+package oci
```
fabriziosestito, line 1: this is `manifest_test.go`

cmd/sbomscannerdb/inspect_format.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+
+	"github.com/kubewarden/sbomscanner/internal/sbomscannerdb/oci"
+)
+
+// renderInspectJSON writes the view as indented JSON.
+func renderInspectJSON(w io.Writer, view oci.ManifestView) error {
```
fabriziosestito: We don't need this function/file. Can you please include this logic in the `runInspect` function in `commands.go`?

fabriziosestito (changes requested):

fabriziosestito (approved):

Issue #1313, opened 2026-07-30
Store enrichment feeds as per-file SQLite databases instead of raw data (eg. JSON, CSV)


alegrey91:
Store each enrichment feed as a SQLite database in the OCI artifact, instead of the raw JSON or CSV file.
The worker queries the database per CVE instead of loading and parsing the whole feed in memory.

Motivation:
- The EPSS feed is large and grows. Loading it in memory on every worker does not scale.
- The worker needs one lookup per CVE, not the whole feed.
- One parser in the build step, none in the worker.

Design:
- `sbomscannerdb build` downloads each feed and builds one SQLite database per feed.
- Each database has one table, `entries`, with the CVE id as the key and the feed entry as JSON.
  This keeps the schema the same for every feed.
- Each database is packed as its own tar.gz layer, with the media type `application/vnd.sbomscanner.db.<feed>.v1.sqlite+tar+gzip`.
- The artifact keeps the `:1` tag and the `lastUpdate` and `nextUpdate` annotations on the manifest.
- The worker opens the databases read-only from its local copy.
  It keeps them open until it stops, and replaces them when it pulls a new artifact.
- The build is reproducible. Entries are inserted in sorted order, the journal is off, and the file is compacted with `VACUUM`.
  The same feed files give the same bytes.

alegrey91:
@fabriziosestito can we consider this closed?

PR #1364, opened 2026-09-07
refactor(worker): sbomscannerdb <-> handlers integration

### LINKED ISSUE #1311: Add configurable enrichment database registry/artifact settings to the SBOMScanner Helm chart
Expose Helm chart configuration so operators can point SBOMScanner workers at the sbomscannerdb enrichment vulnerability database OCI artifact, as designed in [RFC 0010](https://github.com/kubewarden/sbomscanner/blob/main/docs/rfc/0010_enrichment_vulnerability_database.md). The worker pulls this artifact (KEV, EPSS, etc.) from a registry to enrich scan results.

Requirements:
- [x] **Configurable artifact reference**: `worker.sbomscannerDBRepository`. Empty by default until the production database is published.
- [x] **Registry authentication**: the worker uses the Docker `config.json` of the pod when present, the same way trivy does. No dedicated chart value.

### LINKED ISSUE #1312: Integrate the enrichment database into the SBOMScanner worker
Implement the worker-side consumption of the sbomscannerdb enrichment vulnerability database OCI artifact, per RFC 0010 (§Worker integration, §Scan Workflow). This is what actually enriches scan results with
  KEV, EPSS, and future feeds.

## Main things to do

  - [x] Pull & cache: lazy first pull at the first scan (no pull at startup) into a persistent, content-addressed local OCI store under the worker run dir; reuse unchanged blobs, fetch only changed layers.
  - [x] Unpack: decompress each tar.gz layer back to its original file, keyed by media type.
  - [x] Lazy freshness check: at scan time, read nextUpdate from the locally cached manifest; only pull when now >= nextUpdate, otherwise use the local content as-is. No background timer.
  - [x] Strict enrichment: when the database is configured and cannot be updated, the scan fails and is retried, the same way it does when trivy-db cannot be downloaded. A report never carries stale or missing data. The RFC is updated accordingly.
  - [x] Parsing adapters: per-format adapters (KEV, EPSS, etc.) exposing a uniform lookup API to the scan path; unknown media types are ignored (backward-compatible).
  - [x] Config: registry reference/tag from the Helm value `worker.sbomscannerDBRepository` (#1311). The local copy lives under the worker run dir; no separate cache directory setting.
  - [ ] Signature verification: verify the cosign signature before unpacking. Follow-up.

fabriziosestito:
## Description

Fix #1311
Fix #1312

This PR integrates the sbomscanner vulnerability database into the worker, the CLI, and the chart.
The database is an OCI artifact that carries the CISA KEV catalog and the FIRST EPSS scores.
The worker pulls it and adds KEV and EPSS data to every vulnerability in a report.

### API

`Vulnerability` gets two optional fields.

- `kev` holds the CISA catalog entry: `dateAdded`, `dueDate`, and `knownRansomwareCampaignUse`.
- `epss` holds `score`, `percentile`, and `date`.

`score` and `percentile` are strings, like `cvss.v3score`, because the API conventions do not allow floats.
`date` is a `metav1.Time` and carries the `score_date` of the EPSS feed.

### CLI `sbomscannerdb`

All commands use one local OCI store under `$XDG_CACHE_HOME/sbomscannerdb`.

- `build` downloads the feeds and packs one layer per feed. `--data-dir` packs local files instead.
- `push` and `pull` copy the artifact between the local store and a registry.
- `export` writes the data files of an artifact from the local store to a directory.
- `list` and `inspect` read the local store or the registry.

The artifact is tagged with its schema version, `:1`, the same way `trivy-db:2` works.

### Worker

The flag `--sbomscanner-db-repository` enables the database. An empty value disables it.
The worker keeps a local copy under `<run-dir>/sbomscannerdb`.

Before a scan, the worker reads the `nextUpdate` annotation of the local artifact.
When it has passed, the worker pulls the artifact and downloads only the blobs that changed.
When the database cannot be updated, the scan fails, the same way it does when trivy-db cannot be downloaded.

### Chart

`worker.sbomscannerDBRepository` renders the worker flag.
It is empty by default, because the production database is not built and published yet.

## Test

Unit tests, handler tests against a fixed test asset, helm unit tests, and an e2e assertion are part of this PR.

## Checklist

- [ ] I have updated the documentation via a pull request in [docs](https://github.com/kubewarden/docs/tree/main/docs/sbom-scanner) repository.


cmd/worker/main.go
```diff
@@ -158,13 +152,10 @@ func main() {
...
-	enrichment.CleanCache(runDir, logger)
-
-	enrichmentStore := setupEnrichmentStore(ctx, enrichmentDBRepository, enrichmentDBCacheDir, runDir, oci.Config{
-		SkipTLSVerify:  enrichmentDBSkipTLS,
-		PlainHTTP:      enrichmentDBPlainHTTP,
-		AllowAnonymous: true,
-	}, logger)
+	var sbomscannerDB *sbomscannerdb.DB
+	if sbomscannerDBRepository != "" {
+		sbomscannerDB = sbomscannerdb.New(sbomscannerDBRepository, runDir, oci.Config{}, logger)
```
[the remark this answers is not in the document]
fabriziosestito: this was not released.

internal/sbomscannerdb/oci/remote.go
flavio, line 1: interacting with a remote registry that uses a self-signed CA requires to disable tls verification. I'm thinking about users mirroring our db into a private registry (think airgap). It would be nice to allow a custom root CA to be injected and used.

Not something I would do as part of this PR, but something worth to be tracked by an issue if you agree with that.

internal/sbomscannerdb/oci/store.go
```diff
@@ -99,3 +128,88 @@ func (s *Store) open() (*orasoci.Store, error) {
...
+	if _, err := layout.Resolve(ctx, ref); err != nil {
+		if errors.Is(err, errdef.ErrNotFound) {
+			return fmt.Errorf("%s not found in local store (run `build` or `pull` first)", ref)
+		}
+		return fmt.Errorf("resolve %s in local store: %w", ref, err)
+	}
+	return nil
+}
+
+// maxDecompressedLayerSize bounds how much a data layer may decompress to (1 GiB).
```
flavio: I would even lower that to 256 MiB

internal/sbomscannerdb/db.go
```diff
@@ -0,0 +1,238 @@
...
+		return nil
+	}
+
+	// On the first call, load what a previous run left in the local store.
+	if db.loadedDigest() == "" {
+		if view, err := db.local.Inspect(ctx, db.ref); err == nil {
+			if err := db.reload(ctx, view); err != nil {
+				return fmt.Errorf("load sbomscanner DB from local store: %w", err)
+			}
+			db.logger.InfoContext(ctx, "loaded sbomscanner DB from local store", "ref", db.ref, "digest", view.Digest, "kev", len(db.kev), "epss", len(db.epss))
```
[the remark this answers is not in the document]
flavio: that's a legitimate concern

internal/sbomscannerdb/db.go
```diff
@@ -0,0 +1,238 @@
...
+// DB is the local copy of the sbomscanner database, indexed in memory by CVE.
+// A nil *DB is disabled: Update does nothing and Lookup returns empty records.
+type DB struct {
+	ref      string
+	cacheDir string
+	local    *oci.Store
+	remote   *oci.Remote
+	logger   *slog.Logger
+
+	mu   sync.RWMutex
```
flavio: I looked into why we are using a mutex. I might be wrong, please correct me.
I've also had some help from fable with the analysis.

Currently all the worker pulls jobs from the NATS queue in a serial way, there's no concurrency.

It's fine to add the mutex in case we start doing things concurrently. However, the current mutex doesn't protect from a race condition:

```
1. if freshUntil() is in the future → return     (RLock, released)
2. Pull from registry                             (no lock, seconds)
3. Inspect local store                            (no lock)
4. if digest unchanged → setHorizon; return       (Lock, released)
5. reload: Export to cacheDir → load → set digest+horizon
                                                  (no lock during Export/parse;
                                                   Lock only for the final field swap)
```
With two handlers A and B calling Update at once when the artifact is past its `nextUpdate`:

```
A: freshUntil() → 09:00 (past)  → proceed
B: freshUntil() → 09:00 (past)  → proceed          ← both pass
A: remote.Pull(...)  ┐
B: remote.Pull(...)  ┘ two full downloads of the same artifact
A: local.Inspect → digest D1 != loaded → reload
B: local.Inspect → digest D1 != loaded → reload    ← B still sees old digest,
                                                     A hasn't stored D1 yet
A: Export writes kev.json, epss.csv to cacheDir ┐
B: Export writes kev.json, epss.csv to cacheDir ┘  ← same paths, concurrently
A: loadKEV/loadEPSS parse files
B: loadKEV/loadEPSS parse files
A: Lock; kev=..., epss=..., digest=D1, horizon=H1; Unlock
B: Lock; kev=..., epss=..., digest=D1, horizon=H1; Unlock
```

On top of wasting bandwidth, the two concurrent updates would write to the same file, which could lead to corruption.

test/crd/storage.sbomscanner.kubewarden.io_nodevulnerabilityreports.yaml
flavio, line 1: not strictly related with this file, but we should also update the example files too

flavio (changes requested):
LGTM, I left some comments

internal/sbomscannerdb/db.go
fabriziosestito: I removed the mutex. This is a leftover of the first implementation and it will go away when we switch to sqlite, in https://github.com/kubewarden/sbomscanner/issues/1313

internal/sbomscannerdb/db.go
```diff
@@ -0,0 +1,210 @@
...
+	if epss, ok := db.epss[cve]; ok {
+		record.EPSS = &epss
+	}
+	return record
+}
+
+// reload exports the artifact in view from the local store and loads its feeds.
+func (db *DB) reload(ctx context.Context, view oci.ManifestView) error {
+	if _, err := db.local.Export(ctx, db.ref, db.cacheDir); err != nil {
+		return fmt.Errorf("export sbomscanner DB: %w", err)
```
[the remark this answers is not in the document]
fabriziosestito: we always carry both layers

internal/sbomscannerdb/db.go
```diff
@@ -0,0 +1,210 @@
...
+		db.logger.WarnContext(ctx, "cannot read the local sbomscanner DB, pulling", "ref", db.ref, "error", err)
+	}
+	stale := err != nil || !time.Now().Before(nextHorizon(view))
+	if stale {
+		if _, err := db.remote.Pull(ctx, db.local, db.ref); err != nil {
+			return fmt.Errorf("pull sbomscanner DB %s: %w", db.ref, err)
+		}
+		if view, err = db.local.Inspect(ctx, db.ref); err != nil {
+			return fmt.Errorf("inspect sbomscanner DB %s: %w", db.ref, err)
+		}
```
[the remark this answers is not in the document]
fabriziosestito: This is wrong. nextUpdate says when to check the registry again. It does not say the data is expired.

internal/sbomscannerdb/oci/remote.go
fabriziosestito: We decided to expose volume mounts for this, which will land as part of the chart harmonization work.

api/storage/v1alpha1/vulnerabilityreport_types.go
```diff
@@ -153,17 +189,13 @@ type Vulnerability struct {
...
+	// KEV is the CISA Known Exploited Vulnerabilities catalog entry for the CVE.
+	// Nil when the CVE is not in the catalog or no KEV data is available.
+	KEV *KEV `json:"kev,omitempty" protobuf:"bytes,17,opt,name=kev"`
 
-	// EPSSPercentile is the EPSS score's percentile rank (0..1), as a string.
-	// Empty when no EPSS data is available.
-	EPSSPercentile string `json:"epssPercentile,omitempty" protobuf:"bytes,19,opt,name=epssPercentile"`
+	// EPSS is the Exploit Prediction Scoring System data for the CVE.
+	// Nil when no EPSS data is available.
+	EPSS *EPSS `json:"epss,omitempty" protobuf:"bytes,18,opt,name=epss"`
```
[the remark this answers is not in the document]
fabriziosestito: it was not shipped

flavio (approved):

PR #1381, opened 2026-09-21
fix: skip cosign signatures and attestations during catalog creation

flavio:
The catalog step treated cosign signatures and attestations as container images. Cosign stores them with the standard image config media type, so the old check let them through. A `.sig` or `.att` tag made the whole ScanJob fail. A cosign bundle index, which has entries without a platform, made the worker panic.

The catalog step now uses an allow list. A manifest is a container image only when it has an image media type, no artifactType, an image config, and one or more layers with an image layer media type. Every other OCI artifact is skipped with a debug log. Index entries without a platform are skipped too.

The tests have been updated to cover for these scenario. I've pushed some cosign signatures into our `/test` assets on `ghcr.io`.


internal/handlers/create_catalog.go
```diff
@@ -476,16 +482,27 @@ func (h *CreateCatalogHandler) multiArchRefToImages(
...
-			m.Platform.Variant,
-			registry.Spec.Platforms) {
+		if !filters.IsPlatformAllowed(m.Platform, registry.Spec.Platforms) {
+			h.logger.DebugContext(ctx, "Skipping index entry, platform not allowed",
+				"ref", ref.Name(),
+				"digest", m.Digest,
+				"platform", m.Platform,
+				"artifactType", m.ArtifactType,
+			)
 			continue
```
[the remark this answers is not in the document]
flavio: The nil-platform case is covered by a unit test in internal/filters/platform_test.go.

internal/handlers/registry/client.go
```diff
@@ -124,38 +135,107 @@ func (c *Client) GetDescriptor(ctx context.Context, ref name.Reference) (*remote
...
+		return false
+	}
+
+	if !manifest.Config.MediaType.IsConfig() {
+		return false
+	}
+
+	if len(manifest.Layers) == 0 {
+		return false
+	}
```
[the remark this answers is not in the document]
flavio: It's a deliberate design choice, there's no need to scan a `scratch` image.

api/v1alpha1/registry_types.go
```diff
@@ -48,6 +48,8 @@ type RegistrySpec struct {
 	CatalogType string `json:"catalogType,omitempty"`
 	// Repositories is the list of the repositories to be scanned
 	// An empty list means all the repositories found in the registry are going to be scanned.
+	// The scanner skips the tags that do not point to a container image. For example, the scanner
+	// skips signatures, attestations, and Helm charts.
```
fabriziosestito: Let's remove the "tags" wording, since it is not correct here:
```suggestion
    // The scanner only catalogs container images. 
    // It skips other OCI artifacts, for example, signatures, attestations, and Helm charts.
```

internal/filters/platform.go
```diff
@@ -3,15 +3,25 @@ package filters
...
 )
 
-// IsPlatformAllowed verify if the platform of the image is allowed by the registry filter.
-func IsPlatformAllowed(os, architecture, variant string, allowedPlatforms []v1alpha1.Platform) bool {
+// IsPlatformAllowed reports if the platform of an image is allowed by the registry filter.
+//
+// A nil platform is not allowed. Image index entries must have a platform. OCI artifacts
+// that reuse the image index format (for example, cosign referrers) have entries without
+// a platform.
+func IsPlatformAllowed(platform *cranev1.Platform, allowedPlatforms []v1alpha1.Platform) bool {
```
fabriziosestito: I don't think we should make this change.

With this approach, we're introducing a dependency on the Go container registry library into an internal package that shouldn't have any reference to it. It also forces the node filter in `nodescanjob` to construct a synthetic `cranev1.Platform` solely for filtering.

I think the `platform == nil` logic could instead be handled directly on the consumer side, since this behavior is specific to the `CreateCatalog` handler consumer.

internal/handlers/create_catalog.go
```diff
@@ -476,16 +482,27 @@ func (h *CreateCatalogHandler) multiArchRefToImages(
 	images := []storagev1alpha1.Image{}
 
 	for _, m := range manifest.Manifests {
-		if !filters.IsPlatformAllowed(
-			m.Platform.OS,
-			m.Platform.Architecture,
-			m.Platform.Variant,
-			registry.Spec.Platforms) {
+		if !filters.IsPlatformAllowed(m.Platform, registry.Spec.Platforms) {
```
fabriziosestito: Add the `platform == nil` check here.

We could also use a more specific log message, such as "platform not found", since this case is distinct.

See: https://github.com/kubewarden/sbomscanner/pull/1381/changes#r4068566373

internal/handlers/create_catalog.go
```diff
@@ -427,14 +429,18 @@ func (h *CreateCatalogHandler) singleArchRefToImages(
 ) ([]storagev1alpha1.Image, error) {
 	imageDetails, err := registryClient.GetImageDetails(ctx, ref, nil)
 	if err != nil {
+		if errors.Is(err, registryclient.ErrNotContainerImage) || errors.Is(err, registryclient.ErrNoPlatform) {
```
fabriziosestito: For readability, I would split the two `if` statements into separate blocks and use different log messages for each case.

internal/controller/nodescanjob_controller.go
```diff
@@ -128,12 +130,11 @@ func (r *NodeScanJobReconciler) validateNodeAgainstConfig(ctx context.Context, j
...
-	if !filters.IsPlatformAllowed(
-		node.Status.NodeInfo.OperatingSystem,
-		node.Status.NodeInfo.Architecture,
-		"",
-		config.Spec.Platforms,
-	) {
+	nodePlatform := &cranev1.Platform{
+		OS:           node.Status.NodeInfo.OperatingSystem,
+		Architecture: node.Status.NodeInfo.Architecture,
+	}
```
fabriziosestito: See: https://github.com/kubewarden/sbomscanner/pull/1381/changes#r4068566373

internal/handlers/utils_test.go
```diff
@@ -66,6 +66,18 @@ const (
...
+
+	// The nginx-signed repository contains a copy of imageRefSingleArch plus the
+	// cosign artifacts that point to it. The image digest is imageDigestSingleArch.
+	// To create the artifacts again, run these commands with a throwaway key pair:
+	//   crane copy ghcr.io/kubewarden/sbomscanner/test-assets/nginx:1.27.1 ghcr.io/kubewarden/sbomscanner/test-assets/nginx-signed:1.27.1
+	//   cosign generate-key-pair
+	//   # cosign v2: tag-based signature (.sig) and attestation (.att)
+	//   cosign sign --key cosign.key --tlog-upload=false -y ghcr.io/kubewarden/sbomscanner/test-assets/nginx-signed@<digest>
+	//   cosign attest --key cosign.key --tlog-upload=false -y --predicate predicate.json --type custom ghcr.io/kubewarden/sbomscanner/test-assets/nginx-signed@<digest>
+	imageRefSingleArchSigned     = "ghcr.io/kubewarden/sbomscanner/test-assets/nginx-signed:1.27.1"
```
fabriziosestito, line 78: It's okay to have these tests here given the current test structure. However, we should fill the gap by adding integration tests for the `internal/handlers/registry` package, as described in https://github.com/kubewarden/sbomscanner/issues/983.

internal/handlers/registry/client_test.go
```diff
@@ -0,0 +1,206 @@
...
+
+import (
+	"testing"
+
+	cranev1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/types"
+	"github.com/stretchr/testify/assert"
+)
+
+const (
```
fabriziosestito, line 11: Unrelated to this PR:

This is a client package, so unit tests aren't sufficient to cover it. We should add integration tests instead. See https://github.com/kubewarden/sbomscanner/issues/983.

fabriziosestito (changes requested):
Thanks!

This is a regression: we were skipping cosign signatures correctly before the error-handling refactor.

internal/filters/platform.go
flavio: fixed with 930ab1a2d01f0a5468d1d515423522e9ef7770de

internal/handlers/create_catalog.go
flavio: fixed with 930ab1a2d01f0a5468d1d515423522e9ef7770de

internal/controller/nodescanjob_controller.go
flavio: fixed with 930ab1a2d01f0a5468d1d515423522e9ef7770de

internal/handlers/registry/client_test.go
flavio: good point, closing this specific comment since we're going to address that with a future PR

internal/handlers/utils_test.go
flavio: good point, closing since this is going to be addressed by a future PR

internal/handlers/create_catalog.go
flavio: fixed with 930ab1a2d01f0a5468d1d515423522e9ef7770de

flavio:
@fabriziosestito ready for another review

fabriziosestito (approved):

